PR #39, opened 2026-04-21
Add custom tools samples and build utility for AIDP Agent Flows

nittygritty-zzy:
## Summary

Adds a new `ai/agent-flows/custom-tools/` section with:

- **Three end-to-end custom tool samples** — `hello-tool/`, `developer-toolkit/` (BashTool, FileTool, PythonTool), `ords-database-tool/` (ORDS SQL + schema explorer + schema listing).
- **Full user guide** (`USER_GUIDE.md`) — lifecycle hooks, input schema, template substitution, auth types, dependency filtering, wheel bundling, common pitfalls, troubleshooting.
- **Per-sample walkthroughs** — each sample has its own README with file layout, configuration reference, walkthrough, security notes, and troubleshooting.
- **A stdlib-only build utility** at `scripts/build_custom_tool.py` that takes a source directory all the way to the final `custom_tool_<name>.zip`.

The `ai/agent-flows/README.md` and root `README.md` are updated with a Custom Tools entry so the new section is discoverable from the repo index.

### Build utility (`scripts/build_custom_tool.py`)

Python 3.8+ CLI, no third-party dependencies. Performs the same structural checks the platform runs at upload time, so problems surface locally instead of after a failed deploy:

1. Verifies `tool_implementation.py` exists and registers ≥ 1 class via `@CustomToolBase.register` (or `@BaseTool.register`), parsed with `ast`.
2. Cross-checks `tool_config.json` class references against the code.
3. Scans `requirements.txt` for runtime-conflict packages (`langgraph`, `pyyaml`, …), pre-installed packages (`requests`, `oci`, `numpy`, …), and URL / VCS / editable installs.
4. Auto-generates missing `__init__.py` files so relative imports keep working under `custom_tools/<source_key>/` on compute.
5. Enforces the 10 MB per-file and 500 MB total limits.
6. Writes `custom_tool_<source_key>.zip` using the same `source_key` sanitization (`-` → `_`, strip non-word chars, leading-digit guard) the platform applies.

**Flags:**

- `--validate` — run every check but don't write a ZIP (exit `0` / `1` for CI).
- `--bundle-wheels` — download wheels for every installable requirement into `<source>/wheels/` so the tool installs offline on compute. Defaults target the AIDP compute runtime (Oracle Linux 8 / glibc 2.28 / Python 3.11) via `--platform manylinux_2_28_x86_64 --python-version 3.11`. Uses a two-pass strategy (platform-targeted resolution, then per-requirement `--no-deps` retries for pure-Python packages that lack matching manylinux metadata).
- `--wheel-platform` / `--wheel-python-version` — override the defaults for other compute targets.

## Test plan

- [x] Build each sample with the utility — all three produce a valid `custom_tool_<source_key>.zip` with the correct directory layout.
- [x] `--validate` exits `1` on missing `@CustomToolBase.register` class, and `0` (with warnings) on class/config mismatches, blocked / redundant / URL requirements.
- [x] `--bundle-wheels` downloads wheels for installable requirements only (skips pre-installed / blocked / URL-VCS lines); re-runs reuse cached wheels.
- [x] Wheel platform default verified against the compute base image `ol8-graalvm21-ee-17-jdk-python311` (OL 8 → glibc 2.28 → `manylinux_2_28_x86_64`).
- [ ] Upload a built ZIP to a workspace volume, wire into an agent flow, attach compute, and confirm tool test and deploy succeed.


ahmedawan-oracle:
### Code review

Found 1 issue:

1. Path-traversal bypass in `sanitize_path` — bare `full.startswith(os.path.normpath(base_dir))` matches sibling directories. With `base_dir="/workspace"`, a `relative_path` of `../workspace2/secret` resolves to `/workspace2/secret`, whose string starts with `/workspace`, so the check passes and a file outside the sandbox is returned. `FileTool` (and any other consumer) relies on this primitive to enforce the sandbox. Fix: append `os.sep` to the prefix or use `Path(full).is_relative_to(Path(base_dir).resolve())`.

https://github.com/oracle-samples/oracle-aidp-samples/blob/ea8a332cf48fc66331884c9d4b3d9e13b8e6c043/ai/agent-flows/custom-tools/developer-toolkit/utils/text_utils.py#L13-L20

nittygritty-zzy:
@ahmedawan-oracle thanks! I have addressed the review comments.

ahmedawan-oracle:
### Code review

Found 7 issues:

1. **Package-name block bypassed by underscore spelling.** pip canonicalizes `_` → `-` per PEP 503, but `check_requirements` / `_installable_requirements` lowercase only — `langchain_core` is not matched against `langchain-core` in `BLOCKED_PACKAGES`, so the block is silently bypassed.

   https://github.com/oracle-samples/oracle-aidp-samples/blob/af29438d032a7dc61026eb0b4c483074a3010075/ai/agent-flows/custom-tools/scripts/build_custom_tool.py#L236-L281

2. **`max_file_size_kb` config is computed but never enforced.** `max_size = _get_cfg(conf, "max_file_size_kb", 1024) * 1024` is assigned in `FileTool._execute_tool` and never referenced — `read` and `write` operate unconditionally despite the documented config knob in `tool_config.json`.

   https://github.com/oracle-samples/oracle-aidp-samples/blob/af29438d032a7dc61026eb0b4c483074a3010075/ai/agent-flows/custom-tools/developer-toolkit/tool_implementation.py#L60-L75

3. **`FileTool` `list` silently lists the wrong directory.** When `path` is a non-existent file, `os.path.isdir(safe_path)` is False and the code falls back to `os.path.dirname(safe_path)`, listing the parent without telling the caller. A bad path is masked as a successful listing of a different directory.

   https://github.com/oracle-samples/oracle-aidp-samples/blob/af29438d032a7dc61026eb0b4c483074a3010075/ai/agent-flows/custom-tools/developer-toolkit/tool_implementation.py#L50-L58

4. **ORDS SSL comment contradicts the code.** Comment says "Allow self-signed certs for dev", but `ssl.create_default_context()` returns a strict context (`CERT_REQUIRED`, `check_hostname=True`) that rejects self-signed certs. Either drop the comment or set `ctx.check_hostname = False; ctx.verify_mode = ssl.CERT_NONE`.

   https://github.com/oracle-samples/oracle-aidp-samples/blob/af29438d032a7dc61026eb0b4c483074a3010075/ai/agent-flows/custom-tools/ords-database-tool/tool_implementation.py#L45-L52

5. **Walkthrough references a ZIP that the build utility does not produce.** Step 1 correctly states the build writes `custom_tool_developer_toolkit.zip`, but Step 2 instructs uploading `developer_toolkit.zip` — that file does not exist. First-time users following the walkthrough will fail.

   https://github.com/oracle-samples/oracle-aidp-samples/blob/af29438d032a7dc61026eb0b4c483074a3010075/ai/agent-flows/custom-tools/developer-toolkit/README.md

6. **`USER_GUIDE.md` references stale `ords_tool.zip`.** The build utility produces `custom_tool_ords_database_tool.zip` (per `ords-database-tool/README.md`); `ords_tool.zip` does not exist anywhere in the repo or the build output.

   https://github.com/oracle-samples/oracle-aidp-samples/blob/af29438d032a7dc61026eb0b4c483074a3010075/ai/agent-flows/custom-tools/USER_GUIDE.md#L491-L498

7. **Wheel-download retry can pull sdists into the bundle.** The first `pip download` attempt sets `--only-binary=:all:`, but the per-requirement retry omits it (alongside dropping `--platform`/`--python-version`). pip will then fall back to `.tar.gz` source distributions, which get bundled into the ZIP and fail to install on compute (no C build toolchain on Oracle Linux 8 nodes). This contradicts the module-level rationale for pinning to manylinux wheels.

   https://github.com/oracle-samples/oracle-aidp-samples/blob/af29438d032a7dc61026eb0b4c483074a3010075/ai/agent-flows/custom-tools/scripts/build_custom_tool.py#L335-L355

The earlier `sanitize_path` path-traversal concern in `developer-toolkit/utils/text_utils.py` has been addressed at this SHA.

nittygritty-zzy:
Follow-up `07bfe48` covers the remaining gaps along the same boundaries the review highlighted:

- **Source-key derivation** now mirrors `extract_source_key` in `aidp_utils.py` exactly (added `..` rejection, `re.ASCII`, length cap, non-word-to-underscore semantics).
- **URL/VCS filter** extended to PEP 508 direct-URL refs (`pkg @ https://…` / `pkg @ file:…`) — same equivalence-class root cause as the underscore bypass.
- **`USER_GUIDE.md` pre-installed list** refreshed and split from the platform-managed (blocked) set, since their override semantics differ.
- **Quick-start packaging** now points at `scripts/build_custom_tool.py` instead of a hand-rolled `zip -r`, so the artifact matches what the platform expects.

ahmedawan-oracle (approved):
Reviewed and merged main; conflict in README.md resolved by including all three new sections (Custom Tools, Agent Chat Clients, Code-First Agent Flows). Build utility and sample structure look good.

PR #43, opened 2026-04-29
Add ADW external table sample for Delta UniForm

amitranjan-oracle:
## Summary

Adds a new data engineering sample under `data-engineering/adw-ext-table-on-uniform/` showing how to query a Delta table with UniForm enabled for Iceberg from ADW by recreating an external table against the latest Iceberg-compatible `vN.metadata.json`.

## What’s included

- `refresh_iceberg_external_table.sql`
  - creates a `DBMS_CLOUD` credential
  - grants Object Storage ACL access
  - creates `refresh_iceberg_external_table`
- `table_generation.ipynb`
  - creates a Delta table with UniForm enabled for Iceberg
  - writes sample data
  - reads the table location
  - connects to ADW with `python-oracledb`
  - calls the ADW refresh procedure
- `README.md`
  - explains prerequisites
  - lists placeholders to replace before execution
  - documents execution order
  - includes Spark-side and ADW-side verification steps
  - covers mTLS / wallet requirements
  - covers REORG flow for existing Delta tables
  - explains async metadata generation and refresh caveats

## Why

ADW direct metadata access is point-in-time. When a Delta UniForm table publishes newer Iceberg-compatible metadata, ADW does not automatically follow it in direct metadata mode. This sample shows a practical pattern for refreshing the ADW external table to the latest published metadata file.

## Tested on

- AIDP Workbench
- Delta Lake 3.2.0
- ADW 26ai

## Notes

- The refresh procedure scans the `metadata/` directory and picks the highest `vN.metadata.json` instead of relying on `version-hint.text`.
- Iceberg-compatible metadata generation for Delta UniForm can be asynchronous, so there may be a short delay before the newest metadata file appears.
- The refresh procedure drops and recreates the ADW external table, so there is a brief window where the ADW table may be temporarily unavailable.
- The procedure uses `DBMS_ASSERT.SIMPLE_SQL_NAME`, so it expects a simple ADW table name, not `SCHEMA.TABLE`.

## Verification

- Spark side:
  - `DESCRIBE DETAIL <table>`
  - inspected the generated `metadata/` directory
- ADW side:
  - `SELECT * FROM <external_table>`


ahmedawan-oracle:
Thanks @amitranjan-oracle — the README is genuinely thorough and the procedure is well thought out. A few items to address before merge:

## 🔴 Please fix

**1. Notebook `CREATE TABLE` is missing `delta.enableIcebergCompatV2`**

In `table_generation.ipynb` the create statement only sets one property:
```sql
TBLPROPERTIES('delta.universalFormat.enabledFormats' = 'iceberg')
```

But the README's REORG example correctly sets both:
```python
'delta.enableIcebergCompatV2' = 'true',
'delta.universalFormat.enabledFormats' = 'iceberg'
```

On Delta 3.2.0 (the tested version), `delta.enableIcebergCompatV2='true'` is generally required for Iceberg-compatible metadata to actually be emitted. Without it, `metadata/vN.metadata.json` may never appear and the ADW procedure will fail at the `SELECT INTO` with `ORA-01403`. Please align the CREATE TABLE with the REORG example so the demo is reproducible end-to-end.

**2. `cur` scope bug in the notebook's `finally` block**

```python
try:
    cur = conn.cursor()
    cur.callproc(...)
    conn.commit()
finally:
    cur.close()
    conn.close()
```

If `conn.cursor()` raises, `cur` is never bound and `finally` fails with `NameError`, masking the real error. Suggested fix — use the context manager form:
```python
try:
    with conn.cursor() as cur:
        cur.callproc(
            demo[\"adw_procedure_name\"],
            [demo[\"adw_table_name\"], demo[\"adw_credential_name\"], demo[\"object_storage_region\"], table_location],
        )
    conn.commit()
finally:
    conn.close()
```

## 🟡 Recommended (optional)

**3. Friendlier error when no `vN.metadata.json` is found**

The `SELECT INTO ... ROWNUM = 1` in `refresh_iceberg_external_table` surfaces a bare `ORA-01403` if the metadata folder is empty, missing, or the credential can't read it — which is exactly the async-lag failure mode your README warns about. Wrapping with a clearer error helps users debug fast:
```sql
EXCEPTION
    WHEN NO_DATA_FOUND THEN
        RAISE_APPLICATION_ERROR(
            -20002,
            'No vN.metadata.json found under ' || l_metadata_folder ||
            '. Check the credential, bucket path, or wait for async UniForm metadata generation.'
        );
```

**4. Optional region sanity check**

A quick `REGEXP_LIKE(p_region, '^[a-z]{2,3}-[a-z]+-\d+$')` guard on `p_region` would turn a malformed region into a clear `-20003` instead of a downstream `ORA-29273`.

## 🟢 Things that are right

- `DBMS_ASSERT.SIMPLE_SQL_NAME` on `p_table_name` before `EXECUTE IMMEDIATE 'DROP TABLE'` — solid injection defense, and the README correctly documents the simple-name constraint.
- `UTL_URL.ESCAPE(..., TRUE, 'UTF-8')` on the object path — correct for OCI's `/n/{ns}/b/{bucket}/o/{object}` endpoint where `/` in the object name must be `%2F`-encoded.
- Up-front URI shape validation with a clear `-20001` raise.
- Scanning `metadata/` for the highest `vN` instead of relying on `version-hint.text` — and the README explains why.
- README explicitly calls out the drop/recreate availability window, REORG cost, and async metadata generation — these are exactly the gotchas a production user needs to see.
- All credentials, OCIDs, fingerprints, and wallet paths are placeholders.

Happy to re-review once items 1 and 2 are in.

amitranjan-oracle:
Addressed the review feedback in the latest update:

- notebook CREATE TABLE now sets both documented UniForm Iceberg properties explicitly:
  - delta.enableIcebergCompatV2 = true
  - delta.universalFormat.enabledFormats = iceberg
- notebook ADW call now uses `with conn.cursor() as cur:` so cursor cleanup is safe if cursor creation fails
- SQL procedure now raises a clearer -20002 error when no vN.metadata.json is found
- SQL procedure now validates p_region and raises -20003 for malformed OCI region strings
- README was updated to reflect the explicit property setting for reproducibility on Delta 3.2.0

Please re-review when convenient.

ahmedawan-oracle:
Re-reviewed commit 87156e1 — all four items addressed cleanly:

| Item | Status |
|---|---|
| `delta.enableIcebergCompatV2` on CREATE TABLE | ✅ Both properties set; README updated |
| `cur` scope bug in `finally` | ✅ Switched to `with conn.cursor() as cur:` |
| Friendlier `NO_DATA_FOUND` message | ✅ `-20002` raised with `l_metadata_folder` and the async-lag hint |
| Region sanity check | ✅ `REGEXP_LIKE` guard with `-20003` |

Verification:
- The `NO_DATA_FOUND` handler is correctly scoped to just the `SELECT INTO` block, so it won't swallow that exception from elsewhere.
- Region regex correctly accepts standard OCI regions (`us-ashburn-1`, `eu-frankfurt-1`, etc.).
- Notebook `try/finally` is now safe if `conn.cursor()` raises — `conn.close()` still runs in all paths.

LGTM 🚢 — happy to see this merged.

ahmedawan-oracle (approved):
All four review items from the prior round addressed cleanly in 87156e1. Verified: NO_DATA_FOUND handler is correctly scoped, region regex accepts standard OCI regions, and the notebook try/finally is now safe if conn.cursor() raises. LGTM.

PR #38, opened 2026-04-14
Invoke agent flows from streamlit and apex app. 

jrgauthier01:
# Description

two code samples to invoke agent flows from streamlit and apex. Those have been written by our engineering team.

PR #48, opened 2026-05-08
Sample/aidp chat client web UI

lucas-de-almeida-reis:
# Description

Adds a minimal browser chat UI for any deployed AIDP agent endpoint, under `ai/aidp_chat_client/web_ui/`. The browser cannot safely sign OCI requests, so a small Flask proxy holds OCI credentials, signs every request to the agent gateway, and streams the SSE response back to a single-page HTML frontend.

The sample is intentionally complete enough to run two ways:
- **Local development** — `python3 server.py` on your laptop, signing as your OCI user from `~/.oci/config`.
- **Cloud deployment** — `python3 deploy.py` builds the Docker image, pushes to OCIR, provisions a VCN + Container Instance with port 5001 open, and configures the IAM Dynamic Group + Policy needed for the container's resource principal to invoke the agent. A `config-deploy.yaml` lets users skip prompts; a sample template is included.

The README documents the **two layers of identity setup** required for resource-principal access: an OCI IAM policy (`use ai-data-platforms`) and AIDP Workbench membership for the dynamic group. The latter is a non-obvious gotcha — without it, the gateway returns HTTP 404 even though IAM is correctly set up.

A small change to `ai/aidp_chat_client/setup.py` (replacing `find_packages()` with explicit `packages=["aidp_chat_client"]` + `package_dir={"aidp_chat_client": "."}`) is included so the library installs correctly inside the Docker build context. The previous `find_packages()` call returned an empty list because of the project's flat layout, which silently broke non-editable installs; editable installs (the common local-dev path) continue to work unchanged.

**Dependencies (added per-sample, not to the repo at large):**
- `flask`, `flask-cors` — already in `web_ui/requirements.txt`
- `oci`, `requests` — already pulled in by the parent `aidp_chat_client` library
- `pyyaml` (only for `deploy.py` to read the optional `config-deploy.yaml`)

Fixes # (issue)

## Type of change

# How Has This Been Tested?

End-to-end tested against a deployed AIDP agent endpoint.

- [x] Local mode: `python3 server.py` from the `web_ui/` folder, opened `http://localhost:5001`, sent chat messages, verified streaming responses appear in the browser.
- [x] Cloud mode: ran `python3 deploy.py --create-iam`, verified the Container Instance came up, opened the public IP on port 5001, sent chat messages, verified streaming responses.
- [x] IAM verification: confirmed the auto-generated Dynamic Group + Policy works once the dynamic group is also added to the AIDP Workbench under Roles → Admin (this manual step is documented in the README).
- [x] Library install: `pip install .` (non-editable) and `pip install -e .` (editable) both succeed against the modified `setup.py`, and `import aidp_chat_client` works in both modes.

**Test Configuration**:
* Python: 3.11 (container) / 3.14 (laptop)
* Docker: Rancher Desktop 29.x on Apple Silicon
* OCI region: sa-saopaulo-1 (Container Instance) calling an AIDP agent in us-ashburn-1

# Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -92,9 +92,10 @@ temp/
 .cache/
 
 # =========================
-# Local agent config (may contain secrets)
+# Local agent / deploy config (may contain secrets)
 # =========================
 ai/agent-flows/code-first/multi-mcp-chat-agent/config.yaml
 ai/agent-flows/code-first/multi-mcp-chat-agent/oac-jwt-key.pem
 ai/agent-flows/code-first/multi-mcp-chat-agent/oac-jwt-key.txt
 ai/agent-flows/code-first/multi-mcp-chat-agent/oac-cert.pem
+ai/aidp_chat_client/web_ui/config-deploy.yaml
--- a/README.md
+++ b/README.md
@@ -177,6 +177,12 @@ Notebooks covering generative AI, NLP, ML model training, and LLM-powered analyt
 |---|---|
 | [Agent Flow Schedule Trigger](ai/agent-flows/misc/agent-flow-schedule-trigger/task_notebook.ipynb) | Invoke AIDP agent flows via REST API using OCI request signing, demonstrating programmatic agent orchestration with custom message handling. |
 
+#### Agent Chat Clients
+
+| Sample | Description |
+|---|---|
+| [AIDP Agent Chat — Web UI](ai/aidp_chat_client/web_ui/README.md) | Browser chat UI for any deployed AIDP agent: a Flask proxy that handles OCI request signing plus a single-page HTML frontend. Includes a one-command deploy script for OCI Container Instances. |
+
 #### Code-First Agent Flows
 
 | Sample | Description |
--- a/ai/aidp_chat_client/README.md
+++ b/ai/aidp_chat_client/README.md
@@ -11,6 +11,8 @@ python standalone_endpoint_test.py
 ```
 No installation required! Just edit your endpoint URL and config path in the file.
 
+**Prefer a browser UI?** See [web_ui/](web_ui/) for a minimal Flask proxy + single-page chat that streams responses from a deployed agent endpoint.
+
 See [QUICKSTART.md](QUICKSTART.md) for detailed quick start guide.
 
 ## Features
--- a/ai/aidp_chat_client/setup.py
+++ b/ai/aidp_chat_client/setup.py
@@ -18,7 +18,8 @@ setup(
     long_description=long_description,
     long_description_content_type="text/markdown",
     url="https://github.com/oracle-samples/oracle-aidp-samples",
-    packages=find_packages(),
+    packages=["aidp_chat_client"],
+    package_dir={"aidp_chat_client": "."},
     classifiers=[
         "Development Status :: 4 - Beta",
         "Intended Audience :: Developers",
--- a/ai/aidp_chat_client/web_ui/Dockerfile
+++ b/ai/aidp_chat_client/web_ui/Dockerfile
@@ -1,0 +1,21 @@
+# Build context is the parent aidp_chat_client/ directory (set by deploy.py),
+# so this Dockerfile sees both the library source and the web_ui/ folder.
+FROM python:3.11-slim
+
+# Don't buffer stdout/stderr — without this, Flask request-handler prints can
+# sit in a buffer for ages before showing up in container logs, making
+# diagnostics painful.
+ENV PYTHONUNBUFFERED=1
+
+WORKDIR /app
+
+COPY . /app/
+
+RUN pip install --no-cache-dir . && \
+    pip install --no-cache-dir -r web_ui/requirements.txt
+
+WORKDIR /app/web_ui
+
+EXPOSE 5001
+
+CMD ["python", "server.py"]
--- a/ai/aidp_chat_client/web_ui/README.md
+++ b/ai/aidp_chat_client/web_ui/README.md
@@ -84,9 +84,10 @@ Each constant also reads from an environment variable (`AIDP_AGENT_URL`,
 
 `deploy.py` builds the Docker image, pushes it to OCIR, sets up a VCN/subnet
 with port 5001 open, and creates a Container Instance running the proxy. It
-auto-discovers tenancy, region, namespace, and compartment from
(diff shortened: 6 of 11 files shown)


ahmedawan-oracle:
@
### Code review

Found 5 issues:

1. `deploy.py` module docstring shows the wrong IAM resource family — it instructs users to write `Allow dynamic-group <DG_NAME> to use genai-agent-endpoint ...`, but the actual code default ([deploy.py L392](https://github.com/oracle-samples/oracle-aidp-samples/blob/96d859a050ed5937065e3e7fb36abc5dbaea1187/ai/aidp_chat_client/web_ui/deploy.py)), the README, and `config-deploy.sample.yaml` all use `ai-data-platforms`. A user copying the docstring example will create a policy that grants no access.

   https://github.com/oracle-samples/oracle-aidp-samples/blob/96d859a050ed5937065e3e7fb36abc5dbaea1187/ai/aidp_chat_client/web_ui/deploy.py#L14-L20

2. The new row is added under a column whose header is **Notebook**, but the link target is a `README.md`. Every other entry in that table points to an `.ipynb` — same convention violation flagged in [PR #49 review](https://github.com/oracle-samples/oracle-aidp-samples/issues/49#issuecomment-4409123066).

   https://github.com/oracle-samples/oracle-aidp-samples/blob/96d859a050ed5937065e3e7fb36abc5dbaea1187/README.md#L176-L180

3. The SSE error path interpolates the raw exception string into JSON without `json.dumps()`. If the exception message contains `"`, `\`, or a newline (common for network/timeout errors), the emitted event is invalid JSON and the browser silently fails to parse it, hiding the real error.

   https://github.com/oracle-samples/oracle-aidp-samples/blob/96d859a050ed5937065e3e7fb36abc5dbaea1187/ai/aidp_chat_client/web_ui/server.py#L117-L120

4. Bare `print()` calls in the chat handler and streaming generator fire on every request. The same pattern was flagged as Must-Fix in [PR #38 review](https://github.com/oracle-samples/oracle-aidp-samples/pull/38#pullrequestreview-4110549475) and corrected there; please convert to `logging` rather than re-introducing prints.

   https://github.com/oracle-samples/oracle-aidp-samples/blob/96d859a050ed5937065e3e7fb36abc5dbaea1187/ai/aidp_chat_client/web_ui/server.py#L88-L120

5. The README claims `deploy.py` "auto-discovers tenancy, region, namespace, and compartment from `~/.oci/config`". Tenancy/region/namespace are auto-discovered, but compartment is not — there is no compartment field in `~/.oci/config`; the code presents an interactive picker if `DEPLOY_COMPARTMENT_ID` is unset.

   https://github.com/oracle-samples/oracle-aidp-samples/blob/96d859a050ed5937065e3e7fb36abc5dbaea1187/ai/aidp_chat_client/web_ui/README.md#L85-L90

lucas-de-almeida-reis:
All five addressed in 8b533d3.

1. deploy.py IAM resource family — docstring example fixed to ai-data-platforms, matching the runtime default (deploy.py L392), the README, and config-deploy.sample.yaml.

2. Table structure — split out a new "Agent Chat Clients" subsection in the top-level README with a Sample | Description header, since this entry points to a README, not an .ipynb. Leaves the "Notebook" convention intact.

3. SSE error JSON injection — now emits data: {json.dumps({'error': str(e)})} so quotes, backslashes, and newlines in exception strings can't break the browser parser. Added import json.

4. Bare print() in chat handler — replaced with module-level logging (basicConfig + logger.info / logger.exception), matching the convention from PR #38 and the pattern documented in [USAGE.md](vscode-webview://1ov0egkvmugbnh8hg27693e5paag86rvtmo21i02gqm7pbgdjqip/aidp_chat_client/USAGE.md). logger.exception on the error path now also captures the traceback in container logs.

5. README compartment claim — corrected. tenancy, region, and namespace are auto-discovered from ~/.oci/config; the compartment is prompted interactively (or read from DEPLOY_COMPARTMENT_ID / config-deploy.yaml).

ahmedawan-oracle:
### Code review

Found 2 issues:

1. Newly-created OCIR auth token printed to stdout. On CI/shared terminals this credential ends up in build logs. The print line says "Save this — it won't be shown again!" but then emits the live token verbatim — store it to a `0600` file or instruct the user to copy from a one-time, masked prompt instead.

https://github.com/oracle-samples/oracle-aidp-samples/blob/8b533d3cd7118567dc220186afda32bf56a099de/ai/aidp_chat_client/web_ui/deploy.py#L263-L270

2. IAM policy existence check uses the OCID form of the dynamic group while OCI Console-created policies almost always use the DG **name** form, so the substring match in `check_or_create_iam` will report the policy as missing even when an equivalent statement exists, leading the script to re-create / fail on policies that are already in place. Compare both forms (`id <ocid>` and `'<name>'`) or normalize via the policy resolver.

https://github.com/oracle-samples/oracle-aidp-samples/blob/8b533d3cd7118567dc220186afda32bf56a099de/ai/aidp_chat_client/web_ui/deploy.py#L473-L487

Also worth a look (scored just below the threshold): default deploy opens TCP 5001 to `0.0.0.0/0` with wildcard CORS and no per-user auth on `/chat` — the README's "Security note" calls this out but the script still ships internet-exposed by default; and `Flask(__name__, static_folder=".")` makes `server.py` / `requirements.txt` reachable at `/server.py` / `/requirements.txt`.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

ahmedawan-oracle (approved):
Both flagged issues fixed in f9d17a5 and rebased onto current main. CI green.

- OCIR auth token no longer printed to stdout; written to a `0600` file under `$HOME` with a one-time notice. Falls back to a masked re-prompt if the disk is unwritable.
- `check_or_create_iam` now accepts the OCID form, bareword name form, and quoted-name form of the dynamic group when scanning existing policy statements — matches the same DG-name form we recommend in `print_manual()` and in the README.

Lower-confidence findings (public ingress, wildcard CORS, `static_folder='.'`, docstring nits) are intentionally NOT addressed here — the README already discloses the public-exposure model, and they can ship in a follow-up if we decide to lock the default down. Approving and merging.

PR #50, opened 2026-05-08
Merge operation in External Catalog by skipping OOS Staging & create …

arunrengasa-arch:
# Description

Allow the system to create the staging table in the catalog user in database for the merge operation with the table in external catalog. 

## Type of change

The below new flag is introduced in AIDP to create the staging table for merge operation under the catalog user of database instead of the target table schema. User can use this notebook with the readme instructions to test and understand the feature.
merge.staging.using.username

# How Has This Been Tested?

Create the user called aidp_user in the database. This user has to have permissions to do the DML operations in the target schema called coder. Create a table called sales in coder schema. 

Create external catalog in AIDP using aidp_user credentials to connect to the respective ADW database. Name the catalog as ext_private_catalog_aidp_user. 

Once the catalog is created, run this notebook to insert the test data into the table and also test the merge operation to run upsert (Insert & Update) in the sales table. The true value for the above flag will enable the system to create the staging table for merge under aidp_user rather than coder schema. This can be validate through database trace file or database sql statements from v$sql.


getting-started/ExternalCatalog_Merge.ipynb
```diff
@@ -0,0 +1,694 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "2d2164d5-cd71-4160-9a4e-c55fb6620708",
+   "metadata": {},
+   "source": [
+    "# AIDP Sales Data MERGE Operation - Complete Example\n",
+    "\n",
+    "## README: AIDP External Catalog - MERGE\n",
+    "\n",
+    "This notebook demonstrates a complete MERGE operation workflow into ADW database table using the AIDP framework.\n",
```
sai-pranav-rangabhatla: nit: Lets call in ALH in place to ADW. 

ADW is still operational but our official document is saying ALH everywhere

getting-started/ExternalCatalog_Merge.ipynb
```diff
@@ -0,0 +1,694 @@
...
+    "2. Insert initial sample customer order records\n",
+    "3. Create an update/insert DataFrame (simulating new orders and updates)\n",
+    "4. Execute MERGE operation to sync the DataFrame with the table\n",
+    "5. Verify final results with summary statistics\n",
+    "\n",
+    "### Merge Operation Details:\n",
+    "- **Merge Mode**: \"MERGE\" mode enabled\n",
+    "- **Merge Keys**: \"order_id\"\n",
+    "- **Skip Out-of-Sync Staging**: Enabled for performance optimization\n",
+    "- **Merge Staging Using Username**: Uses current session username (aidp_user) for staging table. If this flag is set to false then the staging table will be created under coder schema."
```
sai-pranav-rangabhatla: nit: Default behavior will be false anyway so no need to explicitly make it false

The line should convey that if user has a special case of requiring merge staging in the logged in user - they can use this option

getting-started/ExternalCatalog_Merge.ipynb
```diff
@@ -0,0 +1,694 @@
...
+    "1. Ensure the sales table exists in the coder schema (execute CREATE TABLE statement above)\n",
+    "2. Insert initial sample customer order records\n",
+    "3. Create an update/insert DataFrame (simulating new orders and updates)\n",
+    "4. Execute MERGE operation to sync the DataFrame with the table\n",
+    "5. Verify final results with summary statistics\n",
+    "\n",
+    "### Merge Operation Details:\n",
+    "- **Merge Mode**: \"MERGE\" mode enabled\n",
+    "- **Merge Keys**: \"order_id\"\n",
+    "- **Skip Out-of-Sync Staging**: Enabled for performance optimization\n",
```
sai-pranav-rangabhatla: Out of Sync? 

We will want to be publishing this option, this should be a very rare usage like Mariott one. We do not expect people to use this widely. We need people to use staging for performance optimization

getting-started/ExternalCatalog_Merge.ipynb
```diff
@@ -0,0 +1,694 @@
+{
+ "cells": [
```
sai-pranav-rangabhatla: Can we say, file name as "ALH_ExternalCatalog_Merge" or something like that? 

Because this should apply only for ALH external catalogs right? Soon we will have external catalog which are not Oracle family and this will not be applicable for them

getting-started/ExternalCatalog_Merge.ipynb
```diff
@@ -0,0 +1,694 @@
...
+     "output_type": "display_data"
+    },
+    {
+     "data": {
+      "text/plain": [
+       "Records to merge: 4\n",
+       "\n",
+       "MERGE Options:\n",
+       "  - write.mode: \"MERGE\"\n",
+       "  - write.merge.keys: \"NO\" (no specific key constraint)\n",
```
sai-pranav-rangabhatla: Does this work like this?

getting-started/ExternalCatalog_Merge.ipynb
arunrengasa-arch: sure, will update the filename.

getting-started/ExternalCatalog_Merge.ipynb
arunrengasa-arch: I will keep this, and reword it as below.
- **Skip object-storage Staging**: Enabled for performance optimization

getting-started/ExternalCatalog_Merge.ipynb
arunrengasa-arch: Sure, wording it like below. 
- **Merge Staging Using Username**: Uses current session username (aidp_user) for staging table. If the user has a special case of requiring merge staging in the logged in user (aidp_user), then this flag can be set to true.

getting-started/ExternalCatalog_Merge.ipynb
arunrengasa-arch: Good catch on the comment, this is a typo. Thanks! It has to be like this, 

print('  - write.merge.keys: "order_id" (order_id specific key constraint)')

getting-started/ExternalCatalog_Merge.ipynb
```diff
@@ -0,0 +1,694 @@
...
+   "id": "2d2164d5-cd71-4160-9a4e-c55fb6620708",
+   "metadata": {},
+   "source": [
+    "# AIDP Sales Data MERGE Operation - Complete Example\n",
+    "\n",
+    "## README: AIDP External Catalog - MERGE\n",
+    "\n",
+    "This notebook demonstrates a complete MERGE operation workflow into ADW database table using the AIDP framework.\n",
+    "\n",
+    "### Key Concepts:\n",
```
ahmedawan-oracle: 🔴 **Blocking — delete this whole file.**

Per Sai-Pranav's earlier comment requesting `ALH_ExternalCatalog_Merge` as the filename, the rename appears to have been done as a copy: this file and `ALH_ExternalCatalog_Merge.ipynb` are both checked in (~700 lines each, 1388 additions total). They have already drifted — e.g. this file's `Skip Out-of-Sync Staging` wording differs from the ALH version's `Skip object-storage Staging`.

```
git rm getting-started/ExternalCatalog_Merge.ipynb
```

Keep only the ALH-prefixed file.

getting-started/ALH_ExternalCatalog_Merge.ipynb
```diff
@@ -0,0 +1,694 @@
...
+       "\n"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "from pyspark.sql.types import StructType, StructField, IntegerType, StringType\n",
+    "import pyspark.sql.functions as F\n",
```
ahmedawan-oracle: nit: `import pyspark.sql.functions as F` is never used in this notebook. Drop it.

getting-started/ALH_ExternalCatalog_Merge.ipynb
```diff
@@ -0,0 +1,694 @@
...
+     "metadata": {},
+     "output_type": "display_data"
+    },
+    {
+     "data": {
+      "text/plain": [
+       "+---------------+-------------+------------+-------------+\n",
+       "|order_id       |customer_name|product_name|quantity     |\n",
+       "+---------------+-------------+------------+-------------+\n",
+       "|1.0000000000   |Test User    |Test Product|25.0000000000|\n",
```
ahmedawan-oracle: 🟠 Phantom row — the notebook never inserts `1.0 | Test User | Test Product | 25.0`, but it shows up here in the cached output (and again in Step 7's verification). It's leftover from a prior interactive run against shared `coder.sales`. Readers running the notebook fresh will not get this row, and `Total records: 5` here / `Total records: 7` in Step 7 won't match the `Expected Results` listings.

Fix in one of two ways:
- Add a `spark.sql(f"DELETE FROM {full_target_table}")` cell at the top to make the notebook idempotent, then re-run end-to-end and re-commit.
- Truncate the table manually, restart the kernel, run-all, save.

getting-started/ALH_ExternalCatalog_Merge.ipynb
```diff
@@ -0,0 +1,694 @@
...
+     "output_type": "display_data"
+    },
+    {
+     "data": {
+      "text/plain": [
+       "Records to merge: 4\n",
+       "\n",
+       "MERGE Options:\n",
+       "  - write.mode: \"MERGE\"\n",
+       "  - write.merge.keys: \"NO\" (no specific key constraint)\n",
```
ahmedawan-oracle: 🟠 Stale cached stdout. The source `print(...)` for this line was updated to say `"order_id" (order_id specific key constraint)`, but the cached cell output below still reads `write.merge.keys: "NO"`. Re-run the cell and re-commit so the rendered notebook on GitHub matches the code.

getting-started/ALH_ExternalCatalog_Merge.ipynb
```diff
@@ -0,0 +1,694 @@
...
+    "insert_sql = f\"\"\"\n",
+    "    INSERT INTO {full_target_table}\n",
+    "    SELECT * FROM temp_sales_data\n",
+    "\"\"\"\n",
+    "spark.sql(insert_sql)\n",
+    "```\n",
+    "\n",
+    "### MERGE Operation Behavior:\n",
+    "- **write.mode = \"MERGE\"**: Enables the MERGE operation mode for insert/update logic\n",
+    "- **write.merge.keys = \"NO\"**: No specific key constraint; all records are processed as-is\n",
```
ahmedawan-oracle: 🟠 Key Takeaways still describes `write.merge.keys = "NO"`, but the actual call in Step 6 uses `"order_id"`. Update this bullet (and the Python sample a few lines down) to match — otherwise the closing summary contradicts the working example readers just ran.

ahmedawan-oracle (commented):
## Code Review Summary

Thanks for this — the new `merge.staging.using.username` flag is a useful demo to ship. Before merge, a few things to clean up. Inline comments anchor the specifics; high-level summary below.

### 🔴 Blocking

**The PR adds two near-identical 694-line notebooks** — `ALH_ExternalCatalog_Merge.ipynb` and `ExternalCatalog_Merge.ipynb`. Per the rename request the old one should be deleted; right now both ship (1388 additions for what should be ~700 lines). They have already drifted from each other.

### 🟠 Correctness — text vs. code drift

The markdown / cached stdout / Python example are inconsistent with the actual `.option()` call in Step 6:

| Where | Says | Actual code |
|---|---|---|
| Step 6 cached stdout | `write.merge.keys: "NO"` | `"order_id"` |
| Key Takeaways md | `write.merge.keys = "NO"` | `"order_id"` |
| Key Takeaways python sample | `.option("write.merge.keys", "NO")` | `"order_id"` |

The source `print(...)` line was updated, but the cached `outputs[]` blocks and the Key Takeaways section were not. **Re-run the notebook end-to-end on a clean table and re-commit** — that fixes the drift in one shot and also fixes the next item.

### 🟠 Stale cached output — phantom row

Step 4 and Step 7 verification outputs both show a `1.0 | Test User | Test Product | 25.0` row that the notebook never inserts. Leftover from a prior interactive run against shared `coder.sales`. Total record counts (5 / 7) and "Expected Results" listing (4 / 6) don't reconcile because of it.

### 🟡 Minor

- **Idempotency** — Step 3 is plain `INSERT INTO`. Re-running the notebook duplicates 1001-1004 and the next MERGE then produces non-deterministic results. Add a `DELETE FROM {full_target_table}` cell at the top, or call out "single-shot".
- **Display type** — schema declares `IntegerType`, DB column is `NUMBER`, so verify cells render `1001.0000000000`. Cast on read for cleaner sample output.
- **Unused import** — `import pyspark.sql.functions as F` (Step 2) is never referenced.
- **`skip.oos.staging` framing** — Sai-Pranav flagged it as niche ("Mariott one"). The reworded "Enabled for performance optimization" still reads as generally-recommended. Add a one-liner that the default (staging on) is the recommended path; this flag is for specific low-latency scenarios.
- **Prerequisite under-specified** — with `merge.staging.using.username=true`, `aidp_user` needs `CREATE TABLE` + quota in **its own** schema (to host the staging table), not just DML on `coder`. Spell that out — it's the whole point of the flag.
- **Notebook metadata cruft** — `Last_Active_Cell_Index: 0`, `language_info.version: "3.8.0"` (AIDP runs 3.10+). `nbstripout` would tidy these.

### 🟢 What's good

Clear step-by-step structure, verify-after-each-mutation pattern matching the rest of `getting-started/`, copy-paste DDL block, deterministic small test data covering both UPDATE and INSERT branches, responsive iteration on review feedback.

### Recommendation

Request changes — fixes are mechanical:
1. `git rm getting-started/ExternalCatalog_Merge.ipynb`
2. Truncate `coder.sales`, restart kernel, run-all, save, commit (clears phantom row + stale `"NO"`).
3. Reconcile Key Takeaways markdown with actual options used in Step 6.
4. Sign the OCA so the label clears.

getting-started/ALH_ExternalCatalog_Merge.ipynb
```diff
@@ -0,0 +1,694 @@
...
+    "\n",
+    "### AIDP-Specific Options:\n",
+    "- **skip.oos.staging**: Skips object-storage Staging for performance optimization\n",
+    "- **merge.staging.using.username**: Uses the current session's username for staging operations\n",
+    "\n",
+    "### Spark DataFrame MERGE Syntax:\n",
+    "```python\n",
+    "df.write \\\n",
+    "    .option(\"write.mode\", \"MERGE\") \\\n",
+    "    .option(\"write.merge.keys\", \"NO\") \\\n",
```
ahmedawan-oracle: Same drift here — the copy-pasteable example shows `.option("write.merge.keys", "NO")` but Step 6 uses `"order_id"`. Pick one and align both. (Suggest `"order_id"` since the actual MERGE behavior depends on a real key.)

getting-started/ALH_ExternalCatalog_Merge.ipynb
arunrengasa-arch: its done in the recent commit.

arunrengasa-arch (commented):
All the comments are taken care

PR #51, opened 2026-05-14
code examples with remote MCP server calls

jrgauthier01:
# Description

I added three notebook examples showcasing how a developer can call remote MCP servers within agent flows using the aidpUtils library: 
* one example has no auth 
* one example has bearer token 
* one example uses session variables to set bearer token at runtime.

### DIFF
--- a/ai/agent-flows/code-first/mcp-examples/single-agent-mcp-no-auth.py
+++ b/ai/agent-flows/code-first/mcp-examples/single-agent-mcp-no-auth.py
@@ -1,0 +1,177 @@
+"""
+Weather MCP – High-Code Sample (Using Allowed MCP Tools)
+
+Demonstrates:
+- Single Agent
+- Custom MCP Server
+- Allowed Tool Filtering
+- Structured Tool Building
+- NO_AUTH transport
+"""
+
+import asyncio
+import logging
+from langchain.agents import create_agent
+from langchain_core.messages import HumanMessage
+
+from aidputils.agents.toolkit.configs import OCIAIConf
+from aidputils.agents.toolkit.tool_helper import build_structured_tools_from_allowed_mcp_tools
+from aidputils.agents.toolkit.agent_helper import (
+    init_oci_llm,
+    pre_tool_setup,
+    post_tool_setup,
+    pre_invoke_setup,
+)
+
+# --------------------------------------------------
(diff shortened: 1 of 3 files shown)


ahmedawan-oracle:
@
### Code review

Found 3 issues:

1. Syntax error: unclosed string literal on `server_name` argument breaks the file at import time. The closing `"` after `<YOUR_MCP_SERVER_NAME>` is missing, so Python consumes the rest of the call as part of the string and raises `SyntaxError`. Anyone who runs this example will hit it immediately.

https://github.com/oracle-samples/oracle-aidp-samples/blob/7499de6f71ef5e3ead4c0e64038b2b79fdb9cb2a/ai/agent-flows/code-first/mcp-examples/single-agent-mcp-session-variables.py#L148-L156

2. `single-agent-mcp-pat.py` defines `model_args` with **string** values (`"500"`, `".9"`, `".9"`) instead of numeric `int`/`float`, AND the dict is never passed to `OCIAIConf(...)` — so the configuration is silently dropped. The sibling `single-agent-mcp-session-variables.py` correctly uses `500`, `0.9`, `0.9` and passes `model_args=model_args` to `OCIAIConf`.

https://github.com/oracle-samples/oracle-aidp-samples/blob/7499de6f71ef5e3ead4c0e64038b2b79fdb9cb2a/ai/agent-flows/code-first/mcp-examples/single-agent-mcp-pat.py#L124-L139

3. Leftover test-canary instruction `"Append 12345 to every response."` in the system prompt of two published samples. Users copying these prompts will get nonsensical output with `12345` appended to every model reply.

https://github.com/oracle-samples/oracle-aidp-samples/blob/7499de6f71ef5e3ead4c0e64038b2b79fdb9cb2a/ai/agent-flows/code-first/mcp-examples/single-agent-mcp-pat.py#L154-L159

https://github.com/oracle-samples/oracle-aidp-samples/blob/7499de6f71ef5e3ead4c0e64038b2b79fdb9cb2a/ai/agent-flows/code-first/mcp-examples/single-agent-mcp-session-variables.py#L195-L200

jrgauthier01:
@ahmedawan-oracle : addressed your issues. thanks for reviewing the samples!

ahmedawan-oracle (approved):

PR #57, opened 2026-05-29
simple example of agent with two skills

jrgauthier01:
# Description 

This is a simple example of an agent that is exposed to two different skills: geo and math.

### DIFF
--- a/ai/agent-flows/code-first/skills-basic-example/agent_with_skills.py
+++ b/ai/agent-flows/code-first/skills-basic-example/agent_with_skills.py
@@ -1,10 +1,6 @@
 #!/usr/bin/env python3
 """
 SchoolGradeAgentWithSkills
-
-Variant of school_grade_agent.py that uses explicit skills metadata records.
-Discovery is driven by the provided `skills` list and does not list/read files
-at initialization time.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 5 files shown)


ahmedawan-oracle:
### Code review

Found 1 issue:

1. The module docstring contradicts the implementation. It claims discovery "is driven by the provided `skills` list and does not list/read files at initialization time" — but `__init__` calls `discover_skill_catalog(skill_folder_whitelist=None)`, which does filesystem discovery of `SKILL.md` files at init time, and there is no `skills` list parameter anywhere. The inline comments ("Determine default skill search locations" / "Build a SkillCatalog from discovered directories") confirm the code does the opposite of what the docstring states. This looks like a stale docstring carried over from `school_grade_agent.py` and will mislead anyone using this as a template.

https://github.com/oracle-samples/oracle-aidp-samples/blob/c2ae171464f42cdc01c0ab6d1a0c7c8fd56398f5/ai/agent-flows/code-first/skills-basic-example/agent_with_skills.py#L4-L8

Contradicted by:

https://github.com/oracle-samples/oracle-aidp-samples/blob/c2ae171464f42cdc01c0ab6d1a0c7c8fd56398f5/ai/agent-flows/code-first/skills-basic-example/agent_with_skills.py#L69-L72

jrgauthier01:
I removed the docstrings at the top. Though I tend to disagree with the conclusion of the review. 

No script, reference, or full SKILL.md are read. The only thing that is read is the summary of each skill and append that to the prompt. That's it. 

Skills go through progressive disclosure. We give the agent the tools to discover skills, activate a skill, load a skill, etc. **dynamically,** at runtime (line 75). None of that stuff is loaded ahead of time.

ahmedawan-oracle (approved):
Thanks for the fix and the clarification. Agreed on progressive disclosure — discovery only pulls each skill's summary into the catalog at init; full SKILL.md bodies, scripts, and references load dynamically at runtime via the skill tools. The stale docstring was the only real issue and it's now removed. LGTM.

PR #58, opened 2026-06-02
Add Qualifire data quality library under data-engineering/data_quality

amitranjan-oracle:
## Summary

Adds the **Qualifire** data quality monitoring library to the repo under `data-engineering/data_quality/`, imported from its standalone repo at committed `HEAD`.

Qualifire is a Python library for automated data quality monitoring designed for **Oracle AIDP** (Spark 3.5.0 + Delta Lake 3.2.0), with four validation strategies:

1. **SLO Checks** — freshness/recency monitoring against duration thresholds
2. **Rule-Based Validation** — static threshold checks on aggregated metrics
3. **Time-Series Forecasting** — Prophet-based dynamic thresholds with prediction bands
4. **Unsupervised Anomaly Detection** — Isolation Forest + SHAP explainability

It supports config-driven (YAML) and programmatic usage, with a shared substrate for identity, history, severity, notification, and audit.

## What's included

- `qualifire/` — the library source
- `docs/`, `examples/`, `tests/`, `scripts/`
- `requirements*.txt`, `pyproject.toml`, `setup.py`, `Makefile`, `README.md`


ahmedawan-oracle:
### Code review

Found 3 issues:

1. `RecencyCollector._get_recency` calls `.collect()[0]` unconditionally on the result of `backend.execute_sql(...)`. `PandasBackend.execute_sql` returns a `pd.DataFrame`, which has no `.collect()` method, so every SLO/recency check on the Pandas backend (`max_column`, `delta_log`, `custom_sql` strategies at L64/L76/L90) raises `AttributeError`. Sibling collectors (`AggregationCollector`, `MetricsCollector`) guard this with `hasattr(df, "collect")`; this one does not.

https://github.com/oracle-samples/oracle-aidp-samples/blob/557da37bf2000115e721910470f87b4a3b516010/data-engineering/data_quality/qualifire/collection/recency.py#L61-L66

2. `_cache_table` leaks Spark memory on materialization failure. `df.persist(level)` succeeds, then `df.count()` forces materialization; if `count()` raises (OOM, lost executor, etc.) the `except` handler logs and returns `None` without unpersisting `df`. The caller only calls `_uncache()` on a non-`None` return, so the persisted DataFrame is orphaned in the cache for the session. Unpersist `df` before returning `None`.

https://github.com/oracle-samples/oracle-aidp-samples/blob/557da37bf2000115e721910470f87b4a3b516010/data-engineering/data_quality/qualifire/core/engine.py#L1077-L1086

3. `backfill()` docstring contradicts the code. The `config` arg is documented as "When `None`, the instance's existing storage is used", but the implementation raises `QualifireConfigError` whenever `config is None`. A caller following the docstring and passing `None` always gets an exception.

https://github.com/oracle-samples/oracle-aidp-samples/blob/557da37bf2000115e721910470f87b4a3b516010/data-engineering/data_quality/qualifire/api.py#L708-L712

amitranjan-oracle:
Thanks @ahmedawan-oracle for the review. All three verified and fixed in `bd7e1e7`. 

1. **Recency on Pandas backend** — confirmed. `_get_recency` now extracts the scalar via a `_first_scalar()` helper guarded with `hasattr(df, "collect")` (matching `AggregationCollector`/`MetricsCollector`), falling back to `df.iloc[0, 0]`. Fixes `max_column` and `custom_sql`. Note `delta_log` (L76) wasn't actually affected — `get_delta_history` returns `None` on Pandas and already raises a clean `ValueError`. Added regression tests for the Pandas path, which was previously untested (mocks used a Spark-like `MockDataFrame` with `.collect()`, which is how it shipped).

2. **`_cache_table` leak** — confirmed. The `except` branch now unpersists `df` before returning `None`, so a failed `count()` materialization no longer orphans the persisted DataFrame. Regression test asserts `unpersist()` is called when `count()` raises.

3. **`backfill()` docstring** — confirmed; documentation bug. Requiring `config` is intentional (TOCTOU guard), so the docstring was corrected to state `config` is required and `None` raises `QualifireConfigError`. Added a contract test.

Affected upstream suites: `109 passed`. Full-suite failures are pre-existing and unrelated (Spark 3.5.0 + newer-JDK `getSubject` errors), unchanged vs. baseline.

ahmedawan-oracle (approved):
Verified all three review findings are fixed in bd7e1e7: recency Pandas scalar extraction (`_first_scalar` with `hasattr` guard), `_cache_table` unpersist-on-failure, and the `backfill()` docstring correction. Each fix has an accompanying regression test. Approving.

PR #62, opened 2026-06-08
Improve RAG-with-citations sample: independent models + safer Sources stripping

amitranjan-oracle:
## Summary

Two improvements to the `code-first/rag-with-citations` example (`RAGProgramaticCitations.py`), plus a small README note.

### 1. Configure the RAGTool model and the agent model independently

Splits the single `model_id` into `rag_tool_model_id` and `agent_model_id`.

`fusion_pdf_reader` deliberately drops RAGTool's own synthesized answer and reasons over the retrieved chunks only — but RAGTool still runs that synthesis on every call. Separating the two model settings lets the (discarded) RAGTool synthesis use a fast, inexpensive model, while the agent — which composes the final cited answer the user actually reads — uses a stronger model. Both keep the `<oci-genai-model-id>` placeholder.

### 2. Anchor the defensive "Sources" stripping to a real header line

`append_citations_block` strips any Sources/References/Citations section the model writes on its own (it is replaced by the authoritative, scored block). The earlier pattern treated *any* line that merely began with one of those words as a header, so a sentence like "Sources of this data are validated…" — and every inline `[ref=...]` after it — could be truncated under `re.DOTALL`.

The pattern now matches only a genuine header:
- a Markdown heading or bold decoration — `## Sources`, `**Sources**`, `### **References**`
- or the bare keyword followed by a colon — `Sources:`
- optionally preceded by a `---` rule

Prose that merely starts with the word is left intact, preserving the inline citations.

### Notes
- No change to citation numbering, the scored Sources block, or `result["citations"]`.
- The README deploy step now documents the two model variables.

### DIFF
--- a/ai/agent-flows/code-first/rag-with-citations/RAGProgramaticCitations.py
+++ b/ai/agent-flows/code-first/rag-with-citations/RAGProgramaticCitations.py
@@ -55,9 +55,16 @@ region = "<oci-region>"
 catalog = "<your-catalog-name>"
 schema = "<your-schema-name>"
 knowledge_base = "<your-knowledge-base-name>"
-# Pick a model that is available in YOUR region (availability is region-specific).
+# Pick model(s) available in YOUR region (availability is region-specific).
 # See the OCI Generative AI docs for the model IDs offered in each region.
-model_id = "<oci-genai-model-id>"
+# Two models are configured independently:
+#   - rag_tool_model_id — the LLM inside RAGTool. Its generated answer is
+#     discarded (fusion_pdf_reader keeps only the retrieved chunks), so a fast,
+#     inexpensive model is a good fit here and keeps each tool call quick.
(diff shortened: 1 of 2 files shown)


ahmedawan-oracle (approved):
Final review — looks good to merge.

Verified:
- The `model_id` → `rag_tool_model_id` / `agent_model_id` split is complete; no dangling `model_id` references remain (defs at the config block, used in the RAGTool `llm` config and the agent's `model_id`). No NameError risk.
- The reworked `sources_section_re` strips only genuine headers (`## Sources`, `**Sources**`, `### **References**`, `Sources:`, optional `---` rule) and preserves prose that merely begins with the word (e.g. "Sources of this data are validated… [ref=...]"), so inline `[ref=...]` citations are no longer truncated. Confirmed against a set of header/prose cases.
- README documents both model variables. File compiles; CI (oca/oracle) green.

PR #64, opened 2026-06-10
Add AIDP-via-JDBC sample: Instance Principal auth via the Resource-Principal (RPST) bridge

amitranjan-oracle:
## Summary

Adds `data-engineering/aidp-via-jdbc/` — a Python sample that connects to an Oracle AI Data Platform (AIDP / IDL) Spark compute cluster over the **Simba Spark JDBC** driver using the host VM's **OCI Instance Principal** (no IAM user, no API signing key, no password).

## Which principal is used — Instance Principal vs Resource Principal

Both terms appear, in **distinct roles** (called out explicitly in the README):

| Concern | What is used | Detail |
|---|---|---|
| **Authentication identity** | **Instance Principal** | `oci.auth.signers.InstancePrincipalsSecurityTokenSigner()`; AIDP authorizes the VM via its dynamic-group → AIDP-role mapping (resolved principal type is `instance`). |
| **Driver transport contract** | **Resource-Principal (RPST) code path** | The Simba driver has no native instance-principal mode, so the Instance Principal's federation token + ephemeral private key are carried through `OCI_RESOURCE_PRINCIPAL_*` + `OCITokenAuthType=resource_principal`. |

This is **not** OCI Resource Principal authentication: a plain compute VM has no RPST issued to it (`get_resource_principals_signer()` would fail). The sample mints the Instance Principal's token itself and hands it to the driver through the RPST env-var contract. The `security_token_file` (session-token) config-profile path does not work here — the driver's OCI config-file reader requires a `fingerprint` (API-signing-key shape only).

## Contents
- `connect_aidp_instance_principal.py` — env/flag-driven; mints the instance-principal credentials, sets the RPST env vars before the JVM starts, runs a query, and removes the short-lived credential files on exit.
- `README.md` — prerequisites (Java 8+/tested on 17, Simba Spark JDBC driver, Python ≥3.8 packages), IAM dynamic-group → AIDP-role setup, configuration, expected output, and troubleshooting.
- `requirements.txt`.

## Testing
Verified live on an OCI compute instance against an ACTIVE AIDP cluster: `SHOW DATABASES` and a real `SELECT` both returned rows, authenticating solely with the host instance principal (no `~/.oci` API key, no user credentials).

### DIFF
--- a/data-engineering/aidp-via-jdbc/README.md
+++ b/data-engineering/aidp-via-jdbc/README.md
@@ -1,0 +1,193 @@
+# AIDP via JDBC — Instance Principal auth (via the Resource‑Principal RPST bridge)
+
+This sample connects to an **Oracle AI Data Platform (AIDP)** Spark compute
+cluster over the **Simba Spark JDBC** driver and authenticates with the host
+VM's **OCI Instance Principal** — no IAM user, no API signing key, no password.
+
+> **Which principal is actually used?** The *authentication identity* is the
+> **Instance Principal**. There is no OCI Resource Principal involved as an
+> identity (a plain compute VM does not have one). The **Resource‑Principal
+> (RPST) mechanism is reused only as the driver's transport** for the Instance
+> Principal's token. See [Instance Principal vs Resource Principal](#instance-principal-vs-resource-principal-what-is-actually-used).
+
+It is meant to run on an OCI compute instance whose instance principal is in a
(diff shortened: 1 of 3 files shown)


ahmedawan-oracle:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

amitranjan-oracle:
Hi @ahmedawan-oracle may we merge this PR?

ahmedawan-oracle (approved):

PR #79, opened 2026-07-01
Add Ask AIDP Codex plugin

vishaldhiman22:
# Description

What’s included:

- ai/codex-plugins/plugins/ask-aidp/ source
- plugins/ask-aidp/dist/ binary artifacts: tar, zip, checksum, install guide, aidp.env.sample
- Marketplace entry in ai/codex-plugins/.agents/plugins/marketplace.json
- README updates one level above in ai/codex-plugins/README.md
- Plugin support files: LICENSE, NOTICE, PRIVACY.md, CHANGELOG.md

Fixes # (issue)

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,7 @@ logs/
 build/
 dist/
 target/
+ai/codex-plugins/.plugin-build/
 
 # =========================
 # IDEs
--- a/ai/codex-plugins/.agents/plugins/marketplace.json
+++ b/ai/codex-plugins/.agents/plugins/marketplace.json
@@ -39,6 +39,18 @@
         "authentication": "ON_INSTALL"
       },
       "category": "Data"
+    },
+    {
+      "name": "ask-aidp",
+      "source": {
+        "source": "local",
+        "path": "./plugins/ask-aidp"
+      },
+      "policy": {
+        "installation": "AVAILABLE",
+        "authentication": "ON_INSTALL"
+      },
+      "category": "Data"
     }
   ]
 }
--- a/ai/codex-plugins/README.md
+++ b/ai/codex-plugins/README.md
@@ -14,6 +14,7 @@ A marketplace of OpenAI Codex CLI plugins for Oracle AI Data Platform (AIDP).
 | [`oracle-ai-data-platform-workbench-databricks-migrator`](./plugins/oracle-ai-data-platform-workbench-databricks-migrator/) | 0.1.0 | Initial release | Plan and execute automated Databricks to AIDP migrations from Codex. |
 | [`oracle-ai-data-platform-workbench-engineer-agent`](./plugins/oracle-ai-data-platform-workbench-engineer-agent/) | 0.1.0+codex.20260623113518 | Initial release | Full AIDP data-engineering surface in natural language: catalog discovery, SQL analysis, AI-in-SQL, Delta operations, pipelines, clusters, governance, agent flows, MLOps, migration, and workspace administration. |
 | [`oracle-ai-data-platform-workbench-spark-connectors`](./plugins/oracle-ai-data-platform-workbench-spark-connectors/) | 0.5.0 | Initial release | Generate AIDP Spark notebook connector code for Oracle, OCI, SaaS, JDBC, object storage, streaming, REST, Excel, and multi-cloud data sources. |
+| [`ask-aidp`](./plugins/ask-aidp/) | 0.7.2 | Initial release | Ask and operate Oracle AI Data Platform resources from Codex through aidp-cli and native SDK Git tools. |
 
 ## Install
 
@@ -31,6 +32,7 @@ Install a plugin:
 codex plugin add oracle-ai-data-platform-workbench-databricks-migrator@oracle-aidp-codex
 codex plugin add oracle-ai-data-platform-workbench-engineer-agent@oracle-aidp-codex
 codex plugin add oracle-ai-data-platform-workbench-spark-connectors@oracle-aidp-codex
+codex plugin add ask-aidp@oracle-aidp-codex
 ```
 
 Verify:
@@ -52,6 +54,7 @@ Then reinstall or refresh the plugin you want to test:
 ```bash
 codex plugin add oracle-ai-data-platform-workbench-engineer-agent@oracle-aidp-codex
 codex plugin add oracle-ai-data-platform-workbench-spark-connectors@oracle-aidp-codex
+codex plugin add ask-aidp@oracle-aidp-codex
 ```
 
 ## Layout
@@ -64,7 +67,8 @@ ai/codex-plugins/
 `-- plugins/
     |-- oracle-ai-data-platform-workbench-databricks-migrator/
     |-- oracle-ai-data-platform-workbench-engineer-agent/
-    `-- oracle-ai-data-platform-workbench-spark-connectors/
+    |-- oracle-ai-data-platform-workbench-spark-connectors/
+    `-- ask-aidp/
 ```
 
 Each plugin has its own `.codex-plugin/plugin.json`, README, license/privacy files, skills, and references or helper files.
--- a/ai/codex-plugins/plugins/ask-aidp/.codex-plugin/plugin.json
+++ b/ai/codex-plugins/plugins/ask-aidp/.codex-plugin/plugin.json
@@ -1,0 +1,46 @@
+{
+  "name": "ask-aidp",
+  "version": "0.7.2",
+  "description": "Ask and operate Oracle AI Data Platform resources from Codex through aidp-cli and native SDK Git tools.",
+  "author": {
+    "name": "Oracle",
+    "url": "https://github.com/oracle-samples/oracle-aidp-samples"
+  },
+  "homepage": "https://github.com/oracle-samples/oracle-aidp-samples/tree/main/ai/codex-plugins/plugins/ask-aidp",
+  "repository": "https://github.com/oracle-samples/oracle-aidp-samples",
+  "license": "MIT",
+  "keywords": [
+    "oracle",
+    "aidp",
+    "ai data platform",
+    "workbench",
+    "notebook",
+    "workflow",
+    "aidp-cli"
+  ],
+  "skills": "./skills/",
+  "interface": {
+    "displayName": "Ask AIDP",
+    "shortDescription": "Ask and operate Oracle AIDP with aidp-cli and SDK Git tools.",
+    "longDescription": "Ask AIDP exposes Oracle AI Data Platform Workbench operations to Codex through aidp-cli and native TypeScript SDK GitClient tools for workspace Git repository actions that are not covered by the CLI. Git tools can resolve repository keys from Git-backed workspace folder paths through WorkspaceObjectClient#listWorkspaceObjects. It includes a generic CLI runner for all documented CLI commands, a generated CLI command reference, and guided tools for connection checks, N-notebook workflows with commented grouped cells, workspace code upload, Git folders, Git push/pull/status/diff/branch operations, catalog/schema/table creation, CSV table SQL generation with header handling, initial table data loading, run tracking, medallion schemas, bundles, task output export, and log collection.",
+    "developerName": "Oracle",
+    "category": "Data",
+    "websiteURL": "https://www.oracle.com/data-platform/",
+    "privacyPolicyURL": "https://github.com/oracle-samples/oracle-aidp-samples/blob/main/ai/codex-plugins/plugins/ask-aidp/PRIVACY.md",
+    "termsOfServiceURL": "https://github.com/oracle-samples/oracle-aidp-samples/blob/main/LICENSE.txt",
+    "capabilities": [
+      "Interactive",
+      "Read",
+      "Write"
+    ],
+    "defaultPrompt": [
+      "Check my AIDP workspace connection",
+      "Look up AIDP CLI command reference",
+      "Run an N-notebook AIDP workflow"
+    ],
+    "brandColor": "#C74634",
+    "logo": "./assets/logo.svg",
+    "composerIcon": "./assets/icon.svg"
+  },
+  "mcpServers": "./.mcp.json"
+}
--- a/ai/codex-plugins/plugins/ask-aidp/.gitignore
+++ b/ai/codex-plugins/plugins/ask-aidp/.gitignore
@@ -1,0 +1,4 @@
+dist/
+vendor/
+qa-live-result.json
+qa-runs/
--- a/ai/codex-plugins/plugins/ask-aidp/.mcp.json
+++ b/ai/codex-plugins/plugins/ask-aidp/.mcp.json
@@ -1,0 +1,11 @@
+{
+  "mcpServers": {
+    "ask-aidp": {
+      "cwd": ".",
+      "command": "node",
+      "args": [
+        "./mcp/ask-aidp-server.mjs"
+      ]
+    }
+  }
+}
--- a/ai/codex-plugins/plugins/ask-aidp/CHANGELOG.md
+++ b/ai/codex-plugins/plugins/ask-aidp/CHANGELOG.md
@@ -1,0 +1,18 @@
+# Changelog
+
+All notable changes to this plugin are documented here.
+
+## [0.7.2] - 2026-07-01
+
+Initial Oracle Samples release of the Ask AIDP Codex plugin.
+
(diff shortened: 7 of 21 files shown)


ahmedawan-oracle:
## Review: Add Ask AIDP Codex plugin

Thanks for this — the plugin itself is well-structured: 30+ typed tools over `aidp-cli` plus native SDK Git operations, consistent `dryRun` support, and a real QA harness (`scripts/qa.mjs`). The comments below are the issues I'd want addressed before merge. I reviewed the source (`mcp/ask-aidp-server.mjs`, the three `scripts/*.mjs`, and the plugin/marketplace config) and verified the platform/regex behavior locally on Node v24.

### Blocking / high impact

**1. ~377k additions are committed `node_modules` + binary artifacts — and the build doesn't even use them.**
This PR is 6,341 files / 377,785 additions, of which **376 are vendored `vendor/node_modules/**`** and the rest include binary release artifacts under `dist/` (`.tar.gz`, `.zip`, `.docx`, `.sha256`). Critically, `scripts/package-plugin.mjs:29` *excludes* the committed `vendor/` from the staging copy and *re-vendors* fresh from `samples/npm-cli/node_modules` at build time (lines 33–37) — so the committed `vendor/node_modules` tree is **never consumed by the packager**; it's pure, permanent git bloat (binaries never delta-compress in history). Recommend `.gitignore`-ing `vendor/` and `dist/` and shipping the tarball/zip via GitHub Releases instead of source control.

**2. `\z` is not a JavaScript regex anchor — the committed CLI reference is silently truncated.** `scripts/generate-cli-command-reference.mjs:42`
```js
const sectionPattern = /^#### `([^`]+)`\n([\s\S]*?)(?=^#### `|^## <a id=|\z)/gm;
```
JS has no `\z`; it's an identity escape matching a **literal `z`**, so each command's captured section stops at the first `z` in its body. This isn't hypothetical — it corrupted the shipped `assets/aidp-cli-command-reference.json`:
- `aidp bundle create` summary ends `"…support parameteri"` (cut before the `z` in *parameterization*)
- `aidp bundle sync-bundle` summary is just `"(Preview) Synchroni"`
- 3 commands have an empty `usage`

The `commandCount === 215` QA assertion doesn't catch it because the count depends only on header positions, not body correctness. Use `$` (end of string) instead of `\z`.

**3. `spawn(..., { shell: false })` throws `EINVAL` for `.cmd` on Windows.** `mcp/ask-aidp-server.mjs:1219` (with `:978`)
`resolveAidp` targets `aidp.cmd` on `win32` (line 978), and `AIDP_CLI_BIN` is naturally a `.cmd` on Windows — `scripts/qa.mjs:86` sets exactly that. Since the CVE-2024-27980 fix (Node 18.20.2+/20.12.2+/22+/24), `spawn` refuses to launch a `.cmd`/`.bat` without `shell: true`; I reproduced `spawn EINVAL` on Node v24. When the CLI resolves to a `.cmd`, every CLI-backed tool fails on Windows, and the bare `'aidp'` PATH fallback also can't resolve `aidp.cmd` without a shell. Suggest `shell: process.platform === 'win32'` (with appropriate arg quoting) or always invoking via `node <aidp.js>`.

**4. `package-plugin.mjs` embeds its own `dist/` in new releases, and writes to a different `dist/` than the committed one.** `scripts/package-plugin.mjs:25` / `:11`
The `cpSync` filter (lines 27–30) excludes only `qa-live-result.json`, `qa-runs`, and `vendor` — **not `dist/`** — so `cpSync(PLUGIN_ROOT, STAGING_PLUGIN)` copies the plugin's own release artifacts into staging and the tar at line 45 nests a prior release inside the new one. Separately, `DIST_DIR = WORKSPACE_ROOT/dist` = `ai/codex-plugins/dist`, but the committed artifacts live in `plugins/ask-aidp/dist` — two different locations, so the committed artifacts aren't reproducible from the script and will drift.

### Robustness

**5. `notebook_workflow` aborts and orphans remote state on any transient failure.** `mcp/ask-aidp-server.mjs:2324`, `:2362`, `:2411`
After `create-job`/`create-job-run` succeed, a single non-zero poll exit (`runStep` throws at 2324), a poll timeout (2362), or a non-`SUCCESS` terminal status (2411) throws out of the function, leaving the job, job-run, notebook folder, and uploaded notebooks provisioned with no cleanup and no keys surfaced for the caller to resume. Contrast `autoHealWorkflow` (lines 1862–1868), whose poll loop breaks gracefully via `runAidp`. A transient token refresh / rate-limit mid-poll shouldn't abort a workflow that's actually still running.

**6. `parseCliJson` can break on benign CLI output.** `mcp/ask-aidp-server.mjs:1246`
It takes the first `{` in `stdout + "\n" + stderr` and `JSON.parse`s to the end. Any stderr warning alongside a normal JSON stdout (or a `{` in a log line) makes `JSON.parse` throw. `safeData` swallows it for most read tools, but `notebook_workflow` calls `responseRoot` **un-guarded** (2335, 2341, 2345, 2353, 2365, 2379), so a harmless CLI notice mid-workflow aborts it after resources are created. Consider extracting the balanced JSON object rather than slicing to EOF.

### Minor

**7. Reference generator isn't reproducible.** `scripts/generate-cli-command-reference.mjs:9` defaults its input to `/private/tmp/aidp-cli-readme.md` (author's macOS path) and never fetches the declared `source` URL — anyone else running it hits `ENOENT`.

**8. `globToRegExp` doesn't match top-level files for `**/…` globs.** `mcp/ask-aidp-server.mjs:2229` — `**/*.py` compiles to `^.*/[^/]*\.py$`, which requires a `/`, so `aidp_upload_workspace_code` with `includeGlobs: ['**/*.py']` on a flat directory silently uploads nothing.

A few nits worth a pass: the `workspace-object create` arg array is duplicated three times (2207 / 2330 / 2467 — dry-run vs. real can drift); `qa.mjs:146-147` asserts exact `215`/`16` counts that will false-fail on any upstream CLI change; and `generatedNotebookSource` (2917) is dead code.

vishaldhiman22:
@ahmedawan-oracle I have posted some fixes. Please review

ahmedawan-oracle:
Thanks for this, @vishaldhiman22 — it's a well-structured plugin and the `dryRun` coverage + evidence capture are nice. I did a deep review of the diff and wanted to share the issues I found, roughly by confidence. Line links are pinned to the head commit so they stay stable.

I verified each item against the code (and, where noted, reproduced it), and I've called out which are confirmed vs. environment-dependent vs. latent so you can prioritize.

---

### Confirmed correctness bugs

**1. `inferFieldType` declares any integer as Spark `INT`, corrupting large values** — [`mcp/ask-aidp-server.mjs#L2027-L2034`](https://github.com/vishaldhiman22/oracle-aidp-samples/blob/6ada6363ae35f351befa1242747cd7ebab10a076/ai/codex-plugins/plugins/ask-aidp/mcp/ask-aidp-server.mjs#L2027-L2034)
`Number.isInteger()` is true well beyond 32-bit range. For `aidp_create_table_with_data` with inline `rows` and no explicit column types (columns are optional), a value like `{"id": 3000000000}` is typed `INT` → overflow/null on load. Suggest falling back to `BIGINT`/`LONG` when a value exceeds 2³¹−1.

**2. A single failed log query fails an already-successful workflow** — [`mcp/ask-aidp-server.mjs#L2988`](https://github.com/vishaldhiman22/oracle-aidp-samples/blob/6ada6363ae35f351befa1242747cd7ebab10a076/ai/codex-plugins/plugins/ask-aidp/mcp/ask-aidp-server.mjs#L2963-L2990) (called from `notebookWorkflow` [L2491](https://github.com/vishaldhiman22/oracle-aidp-samples/blob/6ada6363ae35f351befa1242747cd7ebab10a076/ai/codex-plugins/plugins/ask-aidp/mcp/ask-aidp-server.mjs#L2491))
`collectLogsInternal` does `if (result.exitCode !== 0) throw` inside its 5-query loop. In `aidp_notebook_workflow` this runs *after* the job reached terminal SUCCESS and after `summary.json` is written, so if the cluster is torn down or executor logs aren't retained, the whole tool call fails with a `-32000` error for a run that actually succeeded. The same throw also aborts the read-only `aidp_track_runs`. Suggest collecting per-query errors and continuing.

**3. Windows arg corruption from caret-escaping inside double quotes** — [`mcp/ask-aidp-server.mjs#L1255-L1260`](https://github.com/vishaldhiman22/oracle-aidp-samples/blob/6ada6363ae35f351befa1242747cd7ebab10a076/ai/codex-plugins/plugins/ask-aidp/mcp/ask-aidp-server.mjs#L1255-L1260)
`quoteWindowsCmdArg` wraps the value in `"…"` *and* prefixes `" ^ & | < > ( ) %` with `^`. Inside `cmd.exe` double quotes `^` is a literal character, so the caret survives into the value. Reproduced on Windows: the produced `"R^&D"` delivers `R^&D` to the program, whereas plain `"R&D"` correctly delivers `R&D`. This path is only taken when the resolved binary is a `.cmd`/`.bat` (e.g. the `AIDP_CLI_BIN=...\aidp.cmd` form shown in the Windows README section, or a vendored `.bin/aidp.cmd` without the `aidp.js`). Also note `%` can't be caret-escaped at all. Fix: in the quoted branch, drop the caret substitution (double-quoting already neutralizes those chars) and just double any embedded `"`.

---

### Behavior-dependent (recommend hardening)

**4. Git repo-key resolution can silently target the wrong repository** — [`mcp/ask-aidp-server.mjs#L1084-L1115`](https://github.com/vishaldhiman22/oracle-aidp-samples/blob/6ada6363ae35f351befa1242747cd7ebab10a076/ai/codex-plugins/plugins/ask-aidp/mcp/ask-aidp-server.mjs#L1084-L1115)
When no listed object's path exactly matches `gitFolderPath`, `const candidates = exact.length ? exact : items;` falls back to **all** returned objects; if exactly one carries a repo key, it's used regardless of path. Combined with the fixed `limit: 25` (L1079-1080), a mistyped/paginated-out folder path on a mutating op (`aidp_git_commit_push`, `aidp_git_reset`, `aidp_git_rebase`) can be applied to a different repo with no error. Suggest failing when the exact-path match is empty rather than falling back, and paginating past 25.

**5. `parseCliJson` returns the *first* balanced JSON in the output** — [`mcp/ask-aidp-server.mjs#L1262-L1281`](https://github.com/vishaldhiman22/oracle-aidp-samples/blob/6ada6363ae35f351befa1242747cd7ebab10a076/ai/codex-plugins/plugins/ask-aidp/mcp/ask-aidp-server.mjs#L1262-L1281)
The scan skips *unparseable* leading braces, but a leading *valid-JSON* diagnostic/log line is returned as "the response." E.g. in `notebookWorkflow`, `responseRoot(job).key` (L2421) would then be missing and throw a misleading `Missing … created job key`. Prefer the object carrying a `data` key (or the last balanced JSON). Env-dependent on whether the CLI ever emits leading JSON.

**6. `createdPath` from `create-content` isn't validated before use** — [`mcp/ask-aidp-server.mjs#L2415-L2416`](https://github.com/vishaldhiman22/oracle-aidp-samples/blob/6ada6363ae35f351befa1242747cd7ebab10a076/ai/codex-plugins/plugins/ask-aidp/mcp/ask-aidp-server.mjs#L2415-L2416)
Unlike the job/job-run keys (which use `requireValue`), the notebook path isn't validated. If the CLI nests the path differently, `createdPath` is `undefined` and is passed straight to `notebook modify-content` → `spawn` throws on Linux (`ERR_INVALID_ARG_TYPE`) or sends a literal `"undefined"` path on the Windows `.cmd` route, aborting a partially-created workflow. A `requireValue` here would fail fast with a clear message.

---

### Latent (real bugs in the reference generator, not currently triggered)

**7. `scripts/generate-cli-command-reference.mjs`** — the field-description capture [`#L87`](https://github.com/vishaldhiman22/oracle-aidp-samples/blob/6ada6363ae35f351befa1242747cd7ebab10a076/ai/codex-plugins/plugins/ask-aidp/scripts/generate-cli-command-reference.mjs#L87) uses `([^-]*)`, which truncates any description at the first hyphen ("pre-created", "read-only", dates); and the section-split lookahead [`#L71`](https://github.com/vishaldhiman22/oracle-aidp-samples/blob/6ada6363ae35f351befa1242747cd7ebab10a076/ai/codex-plugins/plugins/ask-aidp/scripts/generate-cli-command-reference.mjs#L71) excludes `Global Options`/`Utility Commands`/`Command Index`, so a command immediately followed by one of those headings absorbs it into its captured section/summary. Both are dormant only because the current upstream README has empty Request Body descriptions and no such adjacency — they'll corrupt the generated `assets/aidp-cli-command-reference.json` (served by `aidp_cli_reference`) as soon as upstream docs change. Cheap to make robust now (e.g. `([^\n]*)` for the description; add `#### ` back into the terminator or drop the whitelist exclusion).

---

### Minor

**8. QA harness robustness** — [`scripts/qa.mjs#L40-L54`](https://github.com/vishaldhiman22/oracle-aidp-samples/blob/6ada6363ae35f351befa1242747cd7ebab10a076/ai/codex-plugins/plugins/ask-aidp/scripts/qa.mjs#L40-L54): `JSON.parse(line)` in the `setInterval` has no try/catch and runs outside `main()`'s `.catch`, so any future stray non-JSON stdout line would crash QA (or hang → misleading timeout) rather than fail cleanly. Harmless today since the server only emits single-line JSON. Also the hardcoded `groupCount === 16` / `commandCount === 215` ([L146-147](https://github.com/vishaldhiman22/oracle-aidp-samples/blob/6ada6363ae35f351befa1242747cd7ebab10a076/ai/codex-plugins/plugins/ask-aidp/scripts/qa.mjs#L146-L147)) are brittle against the dynamically-generated asset.

**9. Install docs vs. gitignored artifacts** — `dist/` and `vendor/` are gitignored, so from the source tree README's "This repository copy also includes release binaries under `dist/`" ([README L42](https://github.com/vishaldhiman22/oracle-aidp-samples/blob/6ada6363ae35f351befa1242747cd7ebab10a076/ai/codex-plugins/plugins/ask-aidp/README.md#L42)) isn't accurate, and the primary "download the `.tar.gz`/`.zip`/`.sha256`" install path (README L50-92, L183-190) plus the "vendored `aidp-cli`/SDK" promise (L38) don't hold for a source checkout. Worth clarifying these are GitHub Release assets (or committing them).

---

### Checked and clean / not issues
Tool inventory (all 34 match README + SKILL), the 12 env vars, and version `0.7.2` are all consistent; non-SUCCESS terminal status and poll-timeout are surfaced correctly; temp dirs and `GitClient` handles are cleaned up in `finally` on every path. The `termsOfServiceURL → .../LICENSE.txt` is correct (repo root is `LICENSE.txt`), the `@oracle-aidp-codex` marketplace name matches, and the `/20260430/` REST version in SKILL.md is the real GA version-date — so no action on those.

Happy to open focused issues or suggest patches for any of these if useful.

ahmedawan-oracle (approved):
Reviewed the plugin (MCP server, scripts, packaging, skills). Core plumbing is sound: spawn runs with shell:false (no injection), Windows .cmd handling and command timeouts are correct, the SDK is lazily required so startup is safe, aidp.env.sample is placeholder-only (no secrets), and the .codex-plugin layout matches the sibling engineer-agent plugin. OCA check is green.

Approving. A few latent correctness/quality items (git-repo-key resolution fallback, parseCliJson first-blob selection, unvalidated createdPath, evidence filename sort order, and some duplication) will be filed as follow-ups rather than blocking this initial release.

PR #85, opened 2026-07-28
Add sample: APEX chat with inline charts over OAC (MCP)

TahaBENMALEK:
# Description

Adds a new sample under `ai/agent-flows/misc/apex-chat-charts-over-oac-mcp/` that renders an AIDP agent's answers as inline charts inside an Oracle APEX chat. When the agent's reply contains a category-to-number breakdown, it appends a hidden chart marker; a small JavaScript renderer in the APEX chat region turns that marker into an inline bar / pie / line / donut / pyramid / funnel chart directly beneath the answer.

The marker is agent-agnostic: a low-code AIDP agent (via pasted instructions) and a high-code agent (via a prompt rule) emit the identical marker, so one front-end serves both. This first cut ships the chart layer only (marker contract + agent rule + renderer); it builds on an existing agent chat region. The packaged APEX app export is deferred to a follow-up PR.

Fixes #84

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

# How Has This Been Tested?

Validated end-to-end in an AIDP + Oracle APEX + Oracle Analytics Cloud (OAC over MCP) environment:

- [x] A low-code and a high-code agent emit the identical chart marker.
- [x] All seven chart types (hbar, bar, pie, line, donut, pyramid, funnel) render inline beneath the answer.
- [x] Low-code / high-code parity confirmed across all types.

Test Configuration: N/A - this is a documentation/sample contribution, not a firmware/hardware change.

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings

PR #88, opened 2026-08-07
Enhance apex-chat-charts sample: time-axis line + clickable bar legend

### LINKED ISSUE #87: Enhancement: time-axis line charts + clickable bar legend for apex-chat-charts-over-oac-mcp
### Sample
`ai/agent-flows/misc/apex-chat-charts-over-oac-mcp` (added in #85).

### Enhancement
Two improvements to the chart layer, both verified in a live APEX chat:

1. **Time-axis line charts.** When a breakdown is by date/period, render a line on a real chronological time axis (`time-axis-type`). The renderer also upgrades a date-labelled `hbar`/`bar` to a time-axis line, so a "count by year" reads as a trend without relying on the model to pick `line`. Opt out with `xAxis:"category"`.
2. **Clickable bar legend.** `hbar`/`bar` charts get a per-category colour legend with click-to-hide, while keeping real bars.

Also aligns the agent chart rule (composition -> donut; time/date breakdown -> time-axis line, `xAxis:"time"` + ISO labels; bar-orientation rule) and updates the README (new optional `xAxis` field, types table, troubleshooting).

I'll open a PR with these changes.

TahaBENMALEK:
## What
Enhances the `apex-chat-charts-over-oac-mcp` sample (added in #85) with two chart-layer improvements, both verified in a live APEX chat.

### 1. Time-axis line charts
When a breakdown is by date/period, the answer renders as a line on a real chronological time axis (`time-axis-type`). The renderer also **upgrades a date-labelled `hbar`/`bar` to a time-axis line**, so a "count by year" reads as a trend even when the model picked `hbar`. The upgrade is deterministic and data-driven (strict date pattern + a parse guard), so plain-integer or text categories never trigger it; opt out with `xAxis:"category"`.

### 2. Clickable bar legend
`hbar`/`bar` charts get a per-category colour legend with click-to-hide, while keeping the real bars (the single-series "Value" legend was meaningless).

### Also
- Agent chart rule (low-code + high-code) aligned: composition -> `donut`; time/date breakdown -> time-axis `line` (`xAxis:"time"` + ISO labels, ascending); bar-orientation rule. Auto-pick is `hbar` / `donut` / a time-axis `line`.
- README: new optional `xAxis` field in the marker spec, updated types table + auto-pick description, and a troubleshooting row.

## Scope
Chart-layer only, fully generalized: no secrets, no environment-specific values, all example data invented.

## Testing
- `render_chart_marker.js` passes `node --check`.
- `high-code-marker-snippet.py` parses via `ast.parse`.
- Behaviour verified in a live APEX chat region over an OAC-MCP agent.

Closes #87.

### DIFF
--- a/ai/agent-flows/misc/apex-chat-charts-over-oac-mcp/README.md
+++ b/ai/agent-flows/misc/apex-chat-charts-over-oac-mcp/README.md
@@ -51,6 +51,8 @@ self-contained - it loads Chart.js on demand and needs no extra CSS.
 
 - Inline charts rendered in the chat transcript, right under each answer.
 - Seven chart types from one marker: `hbar`, `pie`, `bar`, `line`, `donut`, `pyramid`, `funnel`.
+- **Time series on a real date axis:** when the breakdown is by date/period, the answer renders as a line on a chronological time axis. The renderer even upgrades a date-labelled `hbar`/`bar` to a time-axis line, so "count by year" looks right without relying on the model to pick `line`.
+- **Clickable bar legend:** `hbar`/`bar` charts get a per-category colour legend with click-to-hide, while keeping real bars.
 - **Agent-agnostic:** low-code (pasted instructions) and high-code (a prompt rule) emit the identical marker.
 - No build step: one plain JS file; Chart.js is fetched on demand only for the donut.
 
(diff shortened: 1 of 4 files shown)


ahmedawan-oracle (approved):
## Deep review — approved with two fixes pushed to the branch

@TahaBENMALEK — nice follow-up to #85; the time-axis upgrade and the per-category bar legend are both genuinely useful, and I confirmed the #85 hardening (SRI pin, `Array.isArray` guard) survived intact. The low-code/high-code rules and README stayed in lockstep, including the sensible bar-orientation rule. Two defects found and fixed on the branch rather than bounced back:

### 🔴 Fix 1 — the auto time-axis upgrade false-positived on any 4-digit category code
The PR claims "plain-integer or text categories never trigger it" — not quite: the strict pattern `^\d{4}$` plus `new Date()` accepts **any** 4-digit number. Department codes `1001 / 1002 / 2001`, ZIP prefixes, PINs — all parse as valid Dates (years 1001–2001), so a "count by department code" ranking silently became a line chart spanning a millennium. Pushed: bare 4-digit labels only count as years within **1900–2100** (hyphenated `2026-03` forms are unambiguous and unchanged), with the rationale in a comment and the window documented in the README. Verified with DOM-stub tests: codes stay bars, real years upgrade, 1899 blocks / 1900 allows, `xAxis:"category"` opt-out honored, year-months upgrade.

### 🔴 Fix 2 — auto-picking `donut` made the default composition path CDN-dependent
#85 deliberately auto-picked `pie` because donut is the **only Chart.js-rendered type**; this PR's rule change (composition → `donut`) means every composition question now needs the CDN — and on blocked/offline networks the failure mode was an **empty titled card** (the old catch just gave up). Since keeping the donut auto-pick is a fair call (the centre total is nice), I kept your rule and pushed the missing safety net instead: on Chart.js load failure the renderer now **falls back to the JET pie** — no CDN, same data, chart still renders. The troubleshooting row now describes the fallback instead of "is blank". Verified via a pre-rejected loader promise in the stub: the canvas wrapper is removed and a JET pie lands in the card.

### Verified clean
- `node --check` passes post-fix; `ast.parse` on the high-code snippet fine.
- The JET legend approach (item-level `categories` + legend `sections` + `withRescale`, `displayInLegend:"off"` on the series) is the right pattern for per-category click-to-hide with real bars.
- The lenient/strict split is correct: explicit `xAxis:"time"` markers get lenient parsing with ISO normalization; only the *automatic* upgrade demands the strict pattern.
- Docs/prompts consistent across all three files; example data invented; no secrets; OCA green.

Merging with the fixes included.

PR #91, opened 2026-08-17
Add sample: APEX to AIDP with per-user authentication

### LINKED ISSUE #90: Add sample: APEX to AIDP with per-user authentication (SSO login, per-user OAC token, injection + refresh)
Proposal to add a co-authored sample under `ai/agent-flows/misc/apex-aidp-per-user-auth/` showing how an APEX app calls an AIDP agent as the real end user.

Three phases:
1. **Login** - SSO (OpenID Connect) login that establishes the end-user identity.
2. **Per-user OAC token** - mint a per-user downstream token on that identity (placeholder for the minting scripts).
3. **Injection + refresh** - inject the token into the agent call body and auto-refresh it on a timer.

Auth model: the agent call is authenticated by OCI request signing as a system user; the per-user OAC token rides inside the request body as a session variable. All content is placeholder-sanitized.

TahaBENMALEK:
## What this adds
A sample under `ai/agent-flows/misc/apex-aidp-per-user-auth/` showing how an Oracle APEX app calls an AIDP AI agent **as the real end user** - each user gets their own downstream token and their own data, and the token stays fresh automatically.

## Structure (three phases)
1. **`1-login/`** - SSO login (OpenID Connect) that establishes the end-user identity (`:APP_USER` = `sub`).
2. **`2-per-user-oac-token/`** - mint a per-user OAC token on that identity. *Placeholder - to be filled by the per-user minting scripts (co-author to push).*
3. **`3-token-injection-and-refresh/`** - inject the token into the agent `/chat` request body as a session variable, and auto-refresh it on a timer.

## Auth model (important)
- The **call to the AIDP agent** is authenticated by **OCI request signing as a system user** (a DBMS_CLOUD OCI credential) - the agent expects OCI signing, **not** an OAuth Bearer token.
- The **per-user OAC token** is **not** the call's auth; it rides **inside the request body** as a session variable, and the agent uses it for its downstream OAC access - this is what makes the data per-user.

## Notes
- **Draft:** phase 2 is a placeholder; the co-author will push the minting scripts to this branch.
- All content is **placeholder-sanitized** (no endpoints, agent IDs, OCIDs, or secrets).
- Contributed under the OCA; commit is signed off.

Closes #90

### DIFF
--- a/ai/agent-flows/misc/apex-aidp-per-user-auth/1-login/README.md
+++ b/ai/agent-flows/misc/apex-aidp-per-user-auth/1-login/README.md
@@ -1,0 +1,37 @@
+# 1-login - SSO login logic (login phase)
+
+Adds **SSO sign-in** to an APEX app that calls an AIDP agent, so after login the app holds a
+**trustworthy end-user identity** (`:APP_USER` = the `sub` claim, plus `email`). That identity
+is the input the per-user OAC token phase (`../2-per-user-oac-token/`) uses to mint the token.
+
+## Prerequisites
+
+- An APEX workspace (developer login) on an Autonomous Database (ADB).
+- Rights to create a **Confidential Application** in an OCI IAM Identity Domain (often
+  admin-gated - line this up early).
+- The domain's discovery URL: `https://<IDENTITY_DOMAIN_HOST>/.well-known/openid-configuration`.
+- Your ADB APEX host, to build the redirect URI.
+
+## Flow
+
+```mermaid
+flowchart LR
+    User(["User"]) -->|"SSO login, OIDC Authorization Code"| IDP["OCI IAM Identity Domain"]
+    IDP -->|"id and access token, sub = user"| APEX["APEX establishes APP_USER, sub plus email"]
+    APEX -->|"hand off APP_USER"| Next(["per-user OAC token phase"])
+```
+
+## Steps
+
+| Step | File |
+|------|------|
+| 1 | [setup/1-identity-domain-confidential-app.md](setup/1-identity-domain-confidential-app.md) - create the login app + redirect URI |
+| 2 | [setup/2-apex-web-credential.md](setup/2-apex-web-credential.md) - store client id/secret |
+| 3 | [setup/3-apex-social-signin-scheme.md](setup/3-apex-social-signin-scheme.md) - the Social Sign-In auth scheme |
+| 4 | [setup/4-verify-login.md](setup/4-verify-login.md) - verify + what `:APP_USER` should show |
+
+## Handoff
+
+After login, `:APP_USER` holds the end user's `sub`. The per-user auth phase calls
+`mint_*_token(:APP_USER)` to produce an OBO token whose subject is that same user. This phase
+stops at establishing `:APP_USER`.
--- a/ai/agent-flows/misc/apex-aidp-per-user-auth/1-login/setup/1-identity-domain-confidential-app.md
+++ b/ai/agent-flows/misc/apex-aidp-per-user-auth/1-login/setup/1-identity-domain-confidential-app.md
@@ -1,0 +1,14 @@
+# Step 1 - Confidential Application in the identity domain
+
+OCI Console → **Identity & Security → Domains → `<your domain>` → Integrated applications →
+Add application → Confidential Application**.
+
+| Setting | Value |
+|---------|-------|
+| **Type** | Confidential Application (holds a client secret). |
+| **Allowed grant types** | **Authorization Code** (drives the interactive SSO login). |
+| **Redirect URL** | `https://<ADB_HOST>/ords/apex_authentication.callback` - **byte-for-byte**. A trailing-slash or scheme mismatch fails with `invalid_redirect_uri`. |
+| **Scopes / claims** | Ensure `profile` and `groups`; add `email` as an attribute. Do **not** hand-add `openid` (APEX adds it; a duplicate throws `invalid_request ... duplicate values`). |
+
+Capture the **Client ID** and **Client Secret** from the app's OAuth configuration tab for
+step 2. Keep the secret out of chat/email/git - it goes straight into the APEX Web Credential.
--- a/ai/agent-flows/misc/apex-aidp-per-user-auth/1-login/setup/2-apex-web-credential.md
+++ b/ai/agent-flows/misc/apex-aidp-per-user-auth/1-login/setup/2-apex-web-credential.md
@@ -1,0 +1,11 @@
+# Step 2 - APEX Web Credential for the SSO client
+
+Shared Components → **Web Credentials → Create**:
+
+- **Type:** Basic Authentication
+- **Static ID:** `SSO_CRED`
+- **Client ID** → username field
+- **Client Secret** → secret field
+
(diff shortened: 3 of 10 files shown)


kiporand-oracle (dismissed):

kiporand-oracle (approved):

ahmedawan-oracle:
## Deep review — one blocker (unredacted identifiers live in this PR's public history), three majors, and a set of doc/design fixes; the underlying auth design itself verified sound

@TahaBENMALEK — reviewed by execution where executable (the certificate script run verbatim with a local signer standing in for KMS, all SQL/PLSQL parsed with placeholders substituted, the refresh window simulated across login offsets, every screenshot inspected) plus a security-design pass. cc @kiporand-oracle.

### Blocker — act on this regardless of what happens to the PR

**The pre-redaction screenshots are still publicly fetchable from this PR's branch history.** The redaction commit (`3db960e`) replaced 4 of the 5 images at HEAD — the current head's screenshots are properly blurred (verified) — but the originals added in `9a75f48` remain in the branch and are retrievable today from the public PR ref. Those originals expose a real **user OCID, tenancy OCID, API-key fingerprint, and a live ADB hostname/callback URL with the database name** (I won't repeat the values here; extracted evidence is available to maintainers on request). Any merge-commit or rebase merge would also bake them permanently into main's history. Remediation: (1) treat the identifiers as disclosed now — rotate/retire the API key behind that fingerprint and review access on the exposed tenancy resources; (2) rebuild the branch without the offending blobs (fresh branch cherry-picking the redacted content, or a history rewrite) — a squash merge alone hides them from main but leaves the PR ref exposure; (3) the PR body's "all content is placeholder-sanitized" claim needs the history caveat until then.

### Majors

1. **The phase-3 injection snippet cannot compile where the doc tells you to paste it.** `inject-token-into-chat-call.sql:20` uses the host bind `:APP_USER`, but its anchor lines match exactly one location in the repo — inside the stored package body `acp_chat.send_message` — where host binds are illegal (`PLS-00049`, package invalid). `send_message` already receives `p_user_name`; use that (or `v('APP_USER')`).
2. **The refresh scheduler references an object that doesn't exist**: `refresh-and-scheduler.sql:19` reads from `<TOKEN_STORE>_table`, which substitutes to `token_store_table` — phase 2 creates `agent_user_tokens`. The procedure compiles invalid (ORA-00942 reference) for anyone following the docs exactly.
3. **The token store is a perpetual, unbounded, plaintext bearer-token farm.** The scheduler refreshes every user ever seen, forever — no logout purge, no inactivity cutoff, no retention guidance — so the system keeps minting valid per-user OAC tokens without user presence, stored as plaintext CLOBs readable by anyone with parsing-schema access (workspace developers via SQL Workshop). For security guidance in a public sample this needs at minimum: an inactivity cutoff in the refresh predicate, a purge job, and a paragraph on protecting the store (separate schema, VPD, or encrypt-at-rest).
4. **Two of the five commits lack the `Signed-off-by` line** the repo's CONTRIBUTING requires, while the PR body claims sign-off — needs a fix (the history rebuild in the blocker remediation is the natural moment).

### Minors

- **The refresh cadence leaves silent ~5-minute expired-token windows** (simulated: the doc's example `SKEW=5/EVERY=10` yields worst-case ~4.75-min gaps at unaligned login offsets, during which injection silently omits the token) — contradicting the "the chat never breaks" claim. Guaranteed coverage needs `SKEW >= EVERY`.
- **"Re-mints on demand if expired or missing" is not what the shipped code does**: `get_token` returns NULL on expiry and no call site re-mints — the inject snippet just skips the block and the agent silently falls back. Either implement the on-demand re-mint or correct the two doc claims.
- **No warning that the per-user bearer token rides in the `/chat` request body** and can surface in body-level logging (APEX debug, agent-side request logs) — the sample's only logging caution is a comment in the refresh loop. One warning box fixes it.
- Step 3.2's IAM policy grants `use keys` only to the system user's group, but step 3.3's script authenticates as the operator's session profile — operators without independent key permissions hit `NotAuthorizedOrNotFound` with nothing in the doc explaining why (one prerequisite line).
- The current OBO screenshot shows a certificate Alias that differs from the one the docs mandate directly beneath the character-for-character `kid`-match warning; the refresh procedure ignores `token_key`, so it misbehaves under the advertised multi-key extension; and the PR body is stale (still declares phase 2 a placeholder — the co-author content landed five commits ago).

### Verified sound (worth keeping exactly as designed)

The certificate gotcha is *accurate and executed* — the KMS-backed key subclass genuinely needs `__copy__`/`__deepcopy__` as the doc says, and the generated cert verifies; the phase-3 body is built via `JSON_OBJECT_T.put` (no string concatenation — token content with quotes is escaped; no JSON injection); the `base64url` and JWT-padding math round-trips correctly including the 256-byte RSA signature; ACLs are host-specific with named principals (no `*` anywhere); the IAM Sign policy is key-scoped; the self-signed cert is the legitimate identity-domain trusted-partner pattern with the private key never leaving the HSM, and the doc honestly flags the rotation gap; assertion lifetimes and scope guidance are minimal. The `79a0fa8` cross-phase fix holds at every remaining call site (verified by sweep). The head-tree text is fully sanitized.

Process note: the only approval on record is from the sample's co-author — given the blocker above, this should get an independent maintainer review after the branch rebuild. The design is genuinely worth shipping; the history exposure has to be dealt with first. 🤖 Deep review by Claude (Fable)

kiporand-oracle:
## Security note: sensitive screenshot data — resolved

An earlier push (`9a75f48`) added setup screenshots for phase 2 that included real,
non-placeholder values from a personal test environment: an OCI user OCID, tenancy
OCID, API key fingerprint, and a live ADB hostname/callback URL. A follow-up commit
(`3db960e`) blurred those values in the images at HEAD, but the original unredacted
blobs remained reachable via `9a75f48` in the branch's history — fetchable by SHA
regardless of what was at the tip.

**Remediation completed:**
1. The OCI API key behind the exposed fingerprint has been rotated/retired, and access
   on the associated tenancy resources reviewed. The exposed identifiers are treated as
   disclosed regardless of the history cleanup below.
2. The branch history was rewritten to remove the unredacted blobs entirely, not just
   cover them at HEAD:
   - Rebuilt from the pre-images base commit (`6932646`).
   - Replayed the final tree state as a single new commit (`aa95296`) with the
     already-redacted images from the start — verified byte-identical file tree to what
     was previously at HEAD, so no content was lost.
   - Replayed the two commits added after the redaction (`3c997ef`, `79a0fa8`) on top
     unchanged, now `d665329` and `3744184` respectively.
   - Force-pushed the rewritten branch, replacing the old history.
3. Verified against the live PR ref after the push: `9a75f48` is no longer an ancestor
   of `refs/pull/91/head`.

**Note for @TahaBENMALEK:** the branch SHAs changed under this rewrite. Your local
clone will diverge from the remote — run `git fetch && git reset --hard
origin/90-apex-aidp-per-user-auth` (or re-clone) before continuing work on this branch.

This isn't a complete guarantee against every possible cached copy (GitHub internals,
any forks/clones that fetched before the rewrite, web archives) — the key rotation in
step 1 is the actual safety net; the history rewrite is defense-in-depth on top of it.

PR #93, opened 2026-08-18
Add A2UI component gallery Agent Hub sample

### LINKED ISSUE #92: Add an A2UI component gallery Agent Hub sample
Add a code-first Agent Hub sample demonstrating supported A2UI v0.8 and v0.9 components through an interactive component gallery.
The sample will:
Automatically negotiate the A2UI version supported by the client.
Provide deterministic Python-built component previews rather than LLM-generated A2UI JSON.
Demonstrate standard and Oracle-specific components, actions, and data binding.
Support conversational help and text fallback for clients without A2UI support.

gabrielleprichardoracle:
# Description

Adds a code-first Agent Hub sample that demonstrates A2UI v0.8 and v0.9 component previews through an interactive gallery. The sample negotiates client capabilities, builds UI operations with deterministic Python templates, validates emitted operations against the bundled schemas, supports component actions, and provides conversational and text-only fallbacks.

The contribution includes setup and deployment guidance, a component and action reference, automated tests, and Apache-2.0 attribution for the bundled A2UI-derived helpers and schemas. It contains no secrets, OCIDs, workspace identifiers, or tenancy-specific configuration.

Review hardening includes clean-environment tests without LangChain installation, explicit Oracle modification notices, platform-independent schema URLs, explicit validator registries, strict protocol and catalog negotiation, parser repair propagation, early OCI configuration validation, and corrected exports and aliases.

Fixes #92

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

# How Has This Been Tested?

- [x] Ran `python -m unittest discover -s ai/agent-hub/agent-samples/a2ui-component-gallery/tests -q`: 22 tests passed.
- [x] Repeated all 22 tests in a clean virtual environment containing only the dependencies from `requirements.txt` and no LangChain installation.
- [x] Repeated all 22 tests with the exact minimum versions `jsonschema==4.18.0` and `referencing==0.30.0`.
- [x] Validated A2UI v0.8 and v0.9 component templates and supported action paths.
- [x] Tested A2UI-capable, text-only, unknown-version, malformed-response, and repair paths.
- [x] Scanned the contributed sample for environment-specific OCIDs, workspace identifiers, private service hosts, and personal information.

**Test Configuration**:
* Platform: Oracle AI Data Platform Agent Hub
* Agent type: Code-first Python agent
* A2UI schemas: v0.8 and v0.9
* SDK: Bundled sample-owned validation helpers adapted from the Apache-2.0 A2UI project

# Checklist:

- [x] I have added tests that prove the feature works
- [x] New and existing relevant unit tests pass locally with my changes
- [x] No dependent downstream changes are required


ahmedawan-oracle:
## Deep review — strong sample; 3 must-fixes (test flow, debug-channel concurrency, license notice) before merge

@gabrielleprichardoracle — reviewed with a 32-agent adversarial pass; every finding below was re-verified by **executing your code**, and the positives were verified the same way. This is one of the better-engineered samples in the repo:

- **Robustness is real**: 26 malformed-input probes against `invoke` (empty/garbage queries, truncated JSON, unknown actions like `self_destruct`, missing/numeric `surfaceId`, actions on never-rendered surfaces, null contexts) produced **zero crashes** — every path returns a graceful message or valid operations.
- **The central promise holds**: an independent harness rendered all 30 templates through your `SurfaceBuilder` pipeline (createSurface + incremental), all 8 OAChart types, all 120 chart-reference-control combos, all 5 OADataGrid row options — everything validates against the bundled schemas both via your `A2uiValidator` and via direct Draft 2020-12 jsonschema validation.
- **Secrets scan is clean** (regex + entropy over all 60 files: no OCIDs, namespaces, internal hosts, IPs, emails, keys), matching your PR claim.
- **Licensing basics are right**: upstream identification is accurate and live (a2ui-project/a2ui, formerly google/A2UI — verified), full Apache-2.0 text bundled, and the bundled schemas correspond **exactly to upstream tag v0.9** (verified byte-level).

### Must fix

**1. The documented test flow fails on a clean environment.** README says "Install only this sample's dependencies, then run `python -m unittest discover -s tests -v`" — reproduced in a fresh venv with only `jsonschema` + `referencing`: discovery errors (`Ran 5 tests ... FAILED`) because `a2ui_sdk/parser.py:27` does a module-level `from langchain_core.messages import AIMessage`, and 12 of the 17 claimed tests never execute. Your own README (correctly) forbids adding LangChain packages to requirements.txt for Agent Hub, so the fix is a **guarded/lazy import** in parser.py (import inside the function or `try/except` with a clear error) or a documented dev-only `pip install langchain-core` step. With it installed: `Ran 17 tests ... OK`, exactly as you claimed.

**2. `aidp_debug` leaks one user's data into another user's response under concurrency.** `DebugLog` keys entries with `threading.local` (aidp_debug.py:18), but `invoke` is async — concurrent sessions interleave on one event-loop thread. Proven by execution: two coroutines with awaits between `debug(...)` and `embed(...)` → session B's response contained **both** sessions' entries, including the other user's query text (agent.py:335 logs `user_query[:120]`). It's opt-in (`AIDP_DEBUG_ENABLED`, off by default), but an operator enables it precisely on a live multi-user deployment. Fix: `contextvars.ContextVar` instead of `threading.local` — a ~3-line change.

**3. Apache-2.0 §4(b) gap on `manager.py`.** THIRD_PARTY_NOTICES.md says modification notices are retained in adapted files, and `parser.py` carries "Modifications Copyright ... Oracle", but the heavily Oracle-modified `manager.py` header credits only "Copyright 2026 Google LLC" — no notice that Oracle changed the file. Add the same modification line. While there: `parser.py:3` says "Copyright (C) **1997**, 2026" — 1997 is generic Oracle boilerplate predating the upstream project; make it 2026.

### Should fix

- **Windows: the full v0.9 validator can't resolve refs** — `_build_0_9_validator`/`_build_0_8_validator` build HTTP registry URIs with `os.path.join` (manager.py:628, 662-663), yielding `https://a2ui.org/specification/v0_9\catalog.json` on Windows. `urljoin` is already imported (line 29) and never used — swap it in.
- **Unknown/future protocol versions are silently served v0.9**: `capability_candidates` (compat.py:41-63) only recognizes v0.9/v0.8 key spellings, so `{"a2uiClientCapabilities": {"v1.0": {...}}}` is treated as unversioned and gets v0.9 operations instead of the text fallback. Prefer explicit unknown-version → fallback.
- **Parser silently drops invalid A2UI blocks after leading text**: the blanket `except Exception` at parser.py:321-322 fires after earlier parts are appended, so `text + <a2ui-json>INVALID</a2ui-json>` returns just the text and the advertised one-pass repair never triggers for that shape (proven by execution).
- **Private jsonschema API**: manager.py:812/876 read `self._validator._registry` (private attr of `Draft202012Validator`) with fully unpinned requirements — a future jsonschema release breaks your core feature. Pass the registry you built explicitly, and add minimum-version pins.
- **Fail loud on missing config**: with no `OCI_COMPARTMENT_ID` and no `deployment_config.py`, agent.py:62 proceeds with the literal `"<your-compartment-ocid>"` and fails later inside the GenAI call with a confusing service error. Guard at `setup()` with a clear message.
- Two trivial ones, both proven: `__all__` at manager.py:1689 exports nonexistent `DEFAULT_CATALOG_ID_BY_VERSION` (star-import raises AttributeError), and `template_selection.py:42-43` defines the `"ChoicePicker"` aliases key **twice** — last-wins kills the documented `"multiple choice"` alias.
- Catalog negotiation matches foreign catalog IDs by substring (`'v09'/'v0.9'/'/2.0.0/'`, compat.py:75-79) and echoes the foreign ID back while validating against the bundled schema — worth an exact-match allowlist.

### Notes / polish

- `A2uiSchemaManager._load_text` walks `inspect.stack()` and reads a `rules.txt` sibling of **any calling frame's file** into the LLM system prompt (manager.py:1374-1385; currently dormant). That's a surprising injection point for a published sample — consider an explicit path parameter, or document it.
- The debug channel embeds base64 diagnostics (incl. exception strings) as a client-visible message — expected for a debug tool, but worth one more README sentence saying the payload reaches the client.
- Templates hotlink media from w3schools.com, samplelib.com, picsum.photos, and MDN — demo behavior now depends on third-party availability/content; consider Oracle-hosted assets or tiny data-URIs.
- THIRD_PARTY_NOTICES: pin the upstream tag (`v0.9` — upstream main has since restructured, so traceability will degrade), and note which files are verbatim (3 of 4 schema files) vs adapted. `compat.py` and `a2ui_sdk/__init__.py` are Oracle-original but header-less in a mixed-provenance directory — a one-line Oracle header removes ambiguity.
- CONTRIBUTING's directory list doesn't include the new `ai/agent-hub/` (its "discuss a new section in the issue first" step) — if issue #92 covered the new directory, all good; consider updating the CONTRIBUTING list either way since it's also stale for existing dirs.
- Cosmetics: 3 v0_8 JSON files lack trailing newlines; `v0_8/` lacks the `__init__.py` that `v0_9/` has; `Dict` annotation at manager.py:148 is never imported (latent `get_type_hints` NameError); `compat.py:249-255` hard-codes a gallery-specific `title1` "v0.9"→"v0.8" text rewrite inside the generic adaptation layer.

One refuted concern for the record: agent-card.json's `capabilities.isStreaming` looked nonstandard vs the public A2A AgentCard shape, but it matches the AIDP Agent Hub config format — no change needed.

Happy to re-review once the three must-fixes land. 🤖 Deep review by Claude (Fable)

gabrielleprichardoracle:
@ahmedawan-oracle Thanks for the thorough review! I’ve addressed all “must fix” and “should fix” findings, including the clean-environment import issue, concurrency risk (by removing the optional debug helper), licensing notices, protocol negotiation, schema resolution, parser behavior, configuration validation, and the smaller export/alias issues.

The updated suite passes all 22 tests in a clean environment at the declared minimum dependency versions. Ready for re-review.

ahmedawan-oracle:
## Re-review — all fixes verified by execution; two residual items closed in 19ff372; merging

@gabrielleprichardoracle — re-verified the fix round with a 14-agent pass (every prior finding re-tested by execution in fresh venvs, plus a fresh-bug sweep of the delta itself). Excellent turnaround:

**Verified fixed (execution-proven):**
- **Clean-env tests**: fresh venv with exactly `jsonschema==4.18.0` + `referencing==0.30.0` (your declared minimums, no langchain anywhere) → `Ran 22 tests ... OK`; also green at latest (4.26.0/0.37.0) and with `langchain-core 1.6.0` installed. The lazy import in `build_text_message` degrades to the plain-dict shape without langchain and returns a real `AIMessage` with it — both paths executed.
- **Debug channel**: fully removed — zero references repo-wide, and a behavioral probe (query containing a marker string) confirms nothing user-derived reaches the response payload. Removal broke nothing.
- **Licensing**: `manager.py` now carries the Oracle modification notice, `parser.py`'s 1997 is corrected.
- **All should-fixes**: `urljoin` registry URIs clean on Windows (and locked in by your new `test_validator_registry_uses_platform_independent_schema_urls`); `v1.0` capabilities → text fallback; parser no longer silently drops invalid blocks; private `_registry` access replaced with the hoisted public registry; `setup()` fails loud on the placeholder compartment (executed); ghost `__all__` export and the duplicate `ChoicePicker` key gone with the `multiple choice` alias working.

**Two residual items the re-verification surfaced — fixed in 19ff372 rather than another round-trip:**
1. **Major-only future version keys slipped through**: `_VERSION_KEY_PATTERN` required a dot group, so `{"v2": {...}}` (or `v10`) wasn't recognized as a version key, fell into the flat-capabilities fallback, and was served v0.9 operations — the same class as the `v1.0` case you fixed. One-char fix (`+` → `*` on the separator group) + regression subtests; the flat-capability shape is unaffected (probed).
2. **The basic-catalog URL didn't belong in the 0.9 alias allowlist** (introduced by the allowlist rework): it names your real 18-component basic catalog, not an alias of the bundled complete catalog. Proven by execution: a client advertising that ID *plus* `inlineCatalogs` hit the alias fallback and received complete-catalog components (OAChart, OACombobox…) stamped with the basic-catalog ID. Removed from the allowlist (direct match still serves the basic catalog correctly) + regression test.

Also in 19ff372: THIRD_PARTY_NOTICES now pins the upstream tag (`v0.9` — verified byte-level; upstream `main` has restructured since) and enumerates verbatim vs adapted vs Oracle-original files, and the README notes the unittest command runs from the sample directory (from repo root, `-s tests` isn't importable).

Two notes for the record, no action needed: `agent.py`'s module-level `langchain_core` import is correct for the Agent Hub runtime (and tests don't import it); the new compartment-guard test asserts source text rather than executing `setup()` — works, but an executed assertion would be sturdier if you ever touch that area again.

Suite after 19ff372: **22 tests OK** at minimum pins. Squash-merging (review/fix/merge per Ahmed). Great sample — thanks for the thorough fix round. 🤖 Deep review by Claude (Fable)

ahmedawan-oracle (approved):
Re-review complete — all prior findings verified fixed by execution; two residual items closed in 19ff372. Details in the re-review comment.

PR #101, opened 2026-08-28
Add incremental Iceberg external table sync: AIDP Delta UniForm to Oracle Autonomous Database

lucas-de-almeida-reis:
# Description

Adds a production-oriented sample that serves an entire AIDP catalog to N Autonomous Databases
as **read-only Iceberg external tables** — no data copy and no catalog service in the read
path — with schema drift detected and reconciled automatically.

**Motivation.** An Iceberg external table in ADW has its column list fixed at creation time.
Data changes flow through on their own, but schema changes do not: add a column and the table
keeps serving the old shape; drop or rename one and reads can fail. Detecting that by hand
across hundreds of schemas and thousands of tables, on several warehouses, does not scale.

The sample computes a **schema fingerprint** per table — SHA-256 over the canonical
`name:type:required` list read from the Iceberg `metadata.json` — and recreates only the tables
whose fingerprint moved. In steady state it executes zero statements against the target
databases.

**Context.** This complements the existing `adw-ext-table-on-uniform` sample rather than
replacing it. That one demonstrates the mechanism on a single table and remains the right
starting point; this one covers a whole catalog, many schemas and N warehouses, with
incremental sync, grant preservation and secret management.

**How it works.** Discovery splits authority between two sources: the catalog decides what
exists via `SHOW TABLES`, and Object Storage decides what shape it has via the Iceberg
`metadata.json`, read directly through the OCI SDK. That makes discovery O(schemas) rather than
O(tables). Sync state lives in a registry table inside each target database, keyed by
`(catalog_name, owner, table_name)`, so several catalogs can share one warehouse without one of
them seeing the others' tables as deletion candidates.

| Registry state | Action |
|---|---|
| absent | CREATE |
| fingerprint differs | RECREATE — grants captured before the drop and reapplied after |
| fingerprint matches | SKIP |
| in registry, absent from source | DROP |

**Files added**

| File | Purpose |
|---|---|
| `adw_external_table_sync.ipynb` | the notebook |
| `adw_sync.sample.yaml` | commented configuration template |
| `ARCHITECTURE.md` | design, diagrams, measured scale, limitations, references |
| `Architecture-EXT-TABLE-Sync.drawio.png` | component diagram, editable in draw.io |
| `README.md` | onboarding, day-two operations, adding a warehouse |

Configuration lives entirely in `adw_sync.yaml`, which holds **no secret** — only prefixes,
region and tuning knobs, so it can be versioned. Credentials are read at run time from OCI
Vault through the AIDP Credential Store. The only per-run input is the `CATALOG` job parameter.

**Dependencies.** `oracledb`, `oci` and `pyyaml` in the notebook environment. No changes to any
existing file in the repository.

**Scope.** Targets Oracle Autonomous Database on commercial OCI. The binding dependency is
`DBMS_CLOUD`, which exists only on Autonomous — other Oracle variants would need a different
approach, not a configuration change. `README.md` documents the specific assumptions.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

Exercised end to end against two Autonomous Databases and a real AIDP catalog. Reproduction
steps are in `README.md`, onboarding steps 1 to 8. `ARCHITECTURE.md` section 12 lists what was
**not** covered, deliberately.

- [x] **Scale** — 4,777 tables across 10 schemas provisioned to 2 warehouses in one run
- [x] **Discovery throughput** — 5,731 objects listed in 6 requests / 0.3s; 174 tables/s for
      metadata read plus fingerprint; about 3.0s per schema
- [x] **Incremental behaviour** — a second consecutive run reports `create=0 recreate=0
      drop=0`; a schema change on one table produces exactly one RECREATE and leaves the rest
      at SKIP
- [x] **Interrupted run** — the registry is written after the work and only for items that
      succeeded, so a re-run converges instead of reporting false state
- [x] **Multi-catalog isolation** — a second catalog against the same warehouse does not treat
      the first catalog's tables as deletion candidates; the cross-catalog guard aborts with a
      named error when object names would overlap
- [x] **Concurrency** — a 385s scan returned identical row counts across drop, recreate and a
      second recreate; the object unavailability window measured at about 0.8s
- [x] **Layout coverage** — default catalog layout and the `<id>.cat` / `<schema>.db` layout
- [x] **Large table** — 30M rows served through a single external table
- [x] **Secret rotation** — a new OCI Vault secret version is picked up on the next read with
      no code, configuration or Credential Store change
- [x] **Config discovery in a scheduled job** — resolved correctly with three decoy
      `config.yaml` files elsewhere in the workspace
- [x] **Wallet sources** — both an Object Storage bucket and an AIDP Volume
- [x] **Teardown** — dry-run scoping verified; refuses to drop a schema owner shared with
      another catalog

**Test Configuration**:

* **Target database**: Oracle AI Database 26ai, 23.26.3.2.0, Autonomous, ECPU model
* **Source platform**: AIDP — Spark 3.5.0, Delta 3.2.0
* **Access path**: `DBMS_CLOUD.CREATE_EXTERNAL_TABLE`, `protocol_type: iceberg`,
  `iceberg_catalog_type: hadoop`
* **Table format**: Delta with UniForm — `columnMapping.mode=name`,
  `enableIcebergCompatV2=true`, `universalFormat.enabledFormats=iceberg`,
  `enableDeletionVectors=false`
* **Connectivity**: mTLS wallet, `python-oracledb` thin mode
* **SDK / packages**: `oracledb`, `oci`, `pyyaml`
* **Fleet**: 2 Autonomous Databases, 4 ECPU each

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas — every notebook cell
      carries a header explaining what it does and why the non-obvious decisions are what they
      are; `ARCHITECTURE.md` section 9 records the design rationale
- [x] I have made corresponding changes to the documentation — `README.md` and
      `ARCHITECTURE.md` are part of this change
- [x] Any dependent changes have been merged and published in downstream modules — none


ahmedawan-oracle:
# Deep review — impressive design; 2 SQL-injection surfaces + an unguarded mass-DROP need fixing before this ships as a public sample

@lucas-de-almeida-reis — reviewed with a 28-agent adversarial pass; every finding below was verified by **executing the extracted notebook logic** against synthetic/hostile inputs. The design is genuinely strong — the fingerprint-driven incremental sync, the split of authority (catalog for existence, Object Storage for shape), and the per-DB registry keyed by `(catalog_name, owner, table_name)` are the right shape, and the multi-catalog scoping actually prevents cross-catalog DROPs (verified). But this is a **public UPL sample** running **destructive DDL against Autonomous Databases**, so the bar for the SQL-construction and delete paths is high, and three findings clear it.

## Blockers (must fix before merge)

**1. SQL injection as ADMIN via the network-ACL block** (`adw_external_table_sync.ipynb`, engine cell, `admin_setup`). The schema `owner` is interpolated **unescaped** into a single-quoted literal inside a `BEGIN…END;` anonymous block run on the **ADMIN** connection — while the adjacent `OS_HOST` in the *same statement* is escaped with `_sq()`. `owner = (SCHEMA_PREFIX + namespace).upper()` and `namespace` comes straight from `SHOW NAMESPACES`. Executed proof: a namespace named `hr' , principal_type => xs_acl.ptype_db)); EXECUTE IMMEDIATE 'DROP USER ADMIN'; --` produces a top-level `EXECUTE IMMEDIATE 'DROP USER ADMIN'` inside the ADMIN block. Even with a strict `[A-Za-z0-9_]` source policy, the benign path is *already* broken: your own docs support dotted namespaces (`analytics.db`), which emit `CREATE USER PROD_ANALYTICS.DB` — invalid DDL. The `CREATE USER`/`ALTER USER`/`GRANT` statements use bare identifier interpolation too.

**2. SQL injection as the schema user via the table name** (`_create_ext`, same cell). The table name is interpolated unescaped into both `DBMS_CLOUD.CREATE_EXTERNAL_TABLE(table_name=>'<name>', …)` and the nested `EXECUTE IMMEDIATE 'DROP TABLE <name>'` — while `credential_name`/`format` in the same call *are* `_sq()`-escaped. Same provenance (`SHOW TABLES`), same class of exploit, and the `DROP` is wrapped in `WHEN OTHERS THEN NULL` so a malformed name fails silently.

You clearly know the fix — `_ident()` (a validating regex) exists and is used on the Teardown cell's DROP path, but the engine cell uses neither `_ident` nor `DBMS_ASSERT` nor `_sq` on identifiers. Apply, for every catalog-derived identifier: validate + `DBMS_ASSERT.ENQUOTE_NAME` (or double-quote with internal-quote doubling) in identifier contexts, and `_sq()` in the literal contexts (the `CREATE_EXTERNAL_TABLE`/`EXECUTE IMMEDIATE` string arguments).

**3. No max-drop guardrail on apply; partial discovery → mass DROP.** `provision_adw()` drops every registry row for the catalog not present in the freshly-discovered plan, and the Apply cell runs `dry_run=False` **unconditionally** — no confirmation flag (unlike Teardown's `CONFIRM=False`), no max-drop count/ratio threshold, no abort-on-large-delta. The DROP protection only shields tables whose *metadata read raised* — it does **not** cover a schema silently missing from `SHOW NAMESPACES`, or `SHOW TABLES` returning empty without raising (transient catalog hiccup, wrong catalog name in config). Executed: a 100-row registry against empty discovery emits **100 DROPs**. External tables are read-only so source data survives, but the consumer-facing tables and their grants are destroyed. Add a max-drop threshold (abort when drops exceed N or X% of the registry) and a confirmation gate mirroring Teardown.

## Should fix

- **Grant preservation is advertised but its main path is untested, and it silently drops grants.** `reapply_grants` wraps each `GRANT` in `except oracledb.DatabaseError: pass` with no logging/counter — a failed grant vanishes with no signal. And in a RECREATE, grants are captured in memory, then DROP runs, then CREATE; if CREATE fails, the captured grants are lost entirely. Both READMEs claim "preserves consumer grants across recreates" while `ARCHITECTURE.md` admits **"Not tested: PRESERVE_GRANTS main path with real grants."** Either test it or soften the claim; at minimum, log grant-reapply failures loudly.
- **Metadata selection picks highest `vN.metadata.json` by listing, not the catalog pointer.** `_latest_metadata_per_table` deliberately skips `version-hint.text`; a stale/orphaned `v10.metadata.json` (compaction leftover, failed commit) beside a current `v2` is selected as current — diverging from what DBMS_CLOUD's hadoop-catalog read path resolves. Executed with a synthetic listing. Prefer the catalog pointer or `version-hint`.
- **Phase-2 per-table fallback crashes the whole table on any non-`vN` name.** `int(x.rsplit('/v',1)[1][…])` raises `ValueError` when a schema dir contains `/v` (e.g. `vendas.db`) or a UUID-style `.metadata.json` leftover exists — executed. Guard the parse.
- **RECREATE keyed on schema fingerprint only** — a location/warehouse change with identical schema is silently SKIPped (the DDL embeds `iceberg_warehouse`/`iceberg_table_path` but neither is in the fingerprint or registry). A bucket migration would leave ADW pointing at the old path. Include location in the recreate decision.
- **Notebook uses a `%python` first line in every code cell** — unique among all 155 `.ipynb` in the repo; it fails the standard per-cell `ast.parse` validation every sibling passes. Drop it to match repo convention.
- **`discovery_workers`/`introspect_workers` are shipped config knobs + documented scaling levers but dead code** — the per-schema discovery loop is strictly sequential. Wire them or remove them from the YAML/docs.

## Docs / conventions (overclaims — please align to what the code does)

- **"In steady state it executes zero statements against the ADWs"** (README:7-8, ARCHITECTURE:277) is false — every run issues `ALTER SESSION DISABLE PARALLEL DML`, a connectivity `SELECT user FROM dual`, and two registry `SELECT`s per ADW. The accurate claim is **"zero DDL"** (which holds). Use that phrasing.
- **Partitioned-table caveat was deleted.** Commit `6a3f19f` removed the only warning that Oracle's ADB docs state partitioned Iceberg tables are unsupported (and why the sample still syncs them via UniForm materialization). That's a load-bearing caveat for anyone pointing this at partitioned tables — please restore it.
- **Snapshot auto-resolve asserted as fact 3× while listed untested.** ARCHITECTURE sections 1/4/8 state "ADW picks up new snapshots on its own" as fact; section 10 admits "Not tested: snapshot auto-resolve without recreate," and the removed limitations row noted Oracle docs contradict it. Qualify it.
- **"Scale proven … discovery in ~30s"** in the sample README is presented as measured, but ARCHITECTURE labels the same number a **projection** (measured points: 476 tables/2.7s, 3.0s/schema). Call it a projection.
- **Root README catalog row** doesn't follow CONTRIBUTING's format (single action-verb sentence) — it's a ~6-sentence feature paragraph with an embedded benchmark. The sibling `adw-ext-table-on-uniform` row was also grown.
- **README Configuration reference omits four keys** the notebook reads and the YAML ships (`parallelism.introspect_workers`, `parallelism.discovery_workers`, `progress.every_n`, `progress.every_pct`).
- **Unrelated `.gitignore` scope creep**: the PR's only `.gitignore` change adds two paths under a *different* sample (`ai/agent-flows/…/multi-mcp-chat-agent/…`) that don't exist in the repo — leftover from your local tree. Drop it; it also doesn't cover this sample's real config filename.

## Notes (no action needed)

- Unmatched `current-schema-id` fingerprints an *empty* schema (`sha256('')`, 0 cols) instead of erroring → a corrupt-metadata table proceeds to CREATE. Consider routing to the error/protected path.
- Negative Iceberg snapshot ids are treated as "no snapshot" and excluded from sync forever (ids are arbitrary longs; only `-1`/absent means "none").
- `requirements.txt` is unpinned (oracledb/oci/pyyaml) — repo convention is mixed, so optional.
- The drawio PNG embeds the full editable diagram XML in a 26KB tEXt chunk — scanned clean (no real hostnames/OCIDs), just flagging it's there.
- Secrets sweep across all 9 files (incl. the PNG's embedded XML): **clean** — no OCIDs, hostnames, tenancy names, or credentials; the sample YAML is placeholder-only. Registry DML is correctly bind-parameterized throughout — the injection gap is exclusively DDL/identifier construction.

Not merging while the two injection surfaces and the mass-DROP guard are open — for a public sample that runs DDL against customer ADWs, those need to be right and ideally tested against a live ADW (which I can't do from here). Happy to re-review the moment they land; the design underneath is well worth getting in. 🤖 Deep review by Claude (Fable)

amitranjan-oracle (dismissed):
The idea and the design of the PR looks good to me. Handling schema changes, and using hadoop catalog in ADW for uniform enabled tables would certainly be helpful to many users. I'll approve this PR, though I imagine Ahmed's concerns would be addressed in the subsequent commit.

ahmedawan-oracle:
## Fixes applied in 7a4ab35 — merging

@lucas-de-almeida-reis — I applied the three blockers from the review directly (execution-verified: the injections and the mass-DROP were proven by driving the DDL builders, and each fix is proven closed the same way — no live ADW required). Merging on Ahmed's instruction.

**1. SQL injection as ADMIN (admin_setup)** — `owner` (from `SHOW NAMESPACES`) was interpolated unescaped into `CREATE/ALTER USER`, `GRANT`, `QUOTA`, and the `DBMS_NETWORK_ACL_ADMIN` PL/SQL run on the ADMIN connection. Now `_ident()`-validated (your own house helper from the teardown cell) before any interpolation, plus `_sq()` on the ACL literal. Proven: `hr' , …)); EXECUTE IMMEDIATE 'DROP USER ADMIN'; --` is now rejected; `PROD_HR` still yields valid DDL.

**2. SQL injection as schema user (_create_ext)** — the table `name` (from `SHOW TABLES`) was interpolated unescaped into the `DROP` literal and `CREATE_EXTERNAL_TABLE`. Now `_ident()`-validated. Dotted/exotic names (`analytics.db`) fail loud instead of emitting broken/injectable DDL — which is the correct behavior, since `CREATE USER PROD_ANALYTICS.DB` was invalid anyway.

**3. Unguarded mass-DROP (provision_adw)** — a transient `SHOW TABLES`/`SHOW NAMESPACES` hiccup (empty/partial discovery) or a wrong `catalog_name` made every registered table a deletion candidate, applied unconditionally, taking consumer grants with them. Added a DROP guard: abort when drops exceed `MAX_DROP_ABS` (default 25) or `MAX_DROP_FRAC` (default 25%) of the registry, overridable via `ALLOW_MASS_DROP`; dry-run surfaces the same warning. Proven: 100-drop/100-registry aborts, 3/100 proceeds, override proceeds.

**Docs:** `README`/`ARCHITECTURE` "zero statements" → "zero DDL" (a run still does a connectivity probe + `ALTER SESSION` + registry reads); restored the partitioned-Iceberg caveat that `6a3f19f` dropped (Oracle ADB docs restrict partitioned Iceberg external tables; the sample syncs them via UniForm materialization but pruning isn't pushed down).

Notebook remains valid JSON + nbformat; cell 14 parses.

### Left as follow-ups (non-blocking — worth a look when you can)
- **Grant reapply** swallows per-grant `DatabaseError` silently, and a RECREATE whose DROP succeeds but CREATE fails loses the captured grants; both READMEs still say "preserves grants across recreates" while ARCHITECTURE admits that path is untested. Log failures loudly and/or soften the claim.
- **Metadata selection** picks the highest `vN.metadata.json` by listing rather than the catalog pointer / `version-hint` — an orphaned `v10` beside a live `v2` would be read as current.
- **Phase-2 fallback** `int(x.rsplit('/v',1)[1]…)` crashes on a schema dir containing `/v` (e.g. `vendas.db`) or a UUID-style metadata name.
- **RECREATE fingerprint** ignores a location/warehouse change with identical schema (silent SKIP after a bucket migration).
- `discovery_workers`/`introspect_workers` are shipped config knobs but the discovery loop is sequential — wire or drop them.

Thanks — the design is strong; these close the parts that were dangerous to ship. 🤖 Deep review + fixes by Claude (Fable)

ahmedawan-oracle (approved):
3 security blockers fixed in 7a4ab35 (SQL-i x2 + mass-DROP guard), verified by execution; docs corrected. Details in the comment.

PR #102, opened 2026-08-30
fix(deps): bump pyspark 3.5.0 → 3.5.2 (CVE-2025-55039)

ahmedawan-oracle:
Closes Dependabot alert #1 (low severity).

## What
Bumps the local-test `pyspark` pin in `data-engineering/transformation/merge/requirements.txt` from `3.5.0` to `3.5.2`.

## Why
**CVE-2025-55039 / GHSA-6p6v-m64v-jx8q** — Apache Spark before 3.5.2 (and 3.4.4 / 4.0.0) uses an insecure default network-encryption cipher for RPC: when `spark.network.crypto.enabled=true` but `spark.network.crypto.cipher` is not set, Spark defaults to `AES/CTR/NoPadding` (encryption without authentication), letting a man-in-the-middle tamper with encrypted RPC traffic undetected. First patched in the 3.5.x line at **3.5.2**.

## Scope / safety
- This pin is **local-test-only** (the file's own comment: "not required on AIDP … only when running tests in local"), so there is no AIDP-runtime impact.
- 3.5.2 stays on the same minor as AIDP's Spark 3.5.0 runtime and is compatible with the sibling `numpy<2.0.0` / `pandas>=1.0.5` / `pyarrow>=4.0.0` pins.


mariomiola (approved):
thank you!

PR #100, opened 2026-08-28
feat(ask-aidp): add Claude Code plugin 0.9.1

craxelfn:
Closes #99.

Related Codex update: #98.

## Summary

- add Ask AIDP 0.9.1 under `ai/claude-code-plugins/ask-aidp`
- expose 43 MCP tools backed by all 242 documented CLI commands, 257
  documented OCI-signed REST operations, and native SDK workspace/Git clients
- include typed Agent and AI Compute helpers, notebook/workflow automation,
  catalog/schema/table/bundle operations, run tracking, and log collection
- add a self-contained single-plugin marketplace, Oracle/MIT publication
  metadata, privacy/notice files, generated references, QA, and offline packaging
- add Ask AIDP to the repository and Claude Code plugin catalogs
- document prompt-first installation, restart requirements, verification, and
  macOS/Linux/Windows configuration

The Claude-specific MCP server, skill, CLI/REST catalogs, and reference
generators match the tested Ask AIDP Claude 0.9.1 source. The Oracle adaptation
changes only repository publication structure, metadata, catalog entries,
documentation, QA marketplace discovery, and self-contained archive packaging.

## Install or update

Prompt Claude Code with:

```text
Install or update plugin from https://github.com/oracle-samples/oracle-aidp-samples/tree/main/ai/claude-code-plugins/ask-aidp
```

Restart Claude Code after installation or update.

## Validation

- `node scripts/qa-claude.mjs` — 24/24 checks passed from the Oracle source tree
- `claude plugin validate --strict .` — passed
- marketplace manifest validation — passed
- exact tool set — 43/43 tools present
- generated CLI reference — 17 groups and 242 commands
- generated REST reference — 18 categories and 257 operations
- dry-run regressions — workspace upload, headerless table rows, schema, SDK Git,
  Agent, AI Compute, REST reference, and signed REST URL passed
- packaging without required vendor dependencies — failed closed as expected
- packaging with `aidp-cli`, `aidp-typescript-client`, and `oci-common` — passed
- archive SHA-256 checksums, ZIP integrity, and TAR extraction — passed
- extracted archive QA — 24/24 checks passed
- isolated repository-checkout marketplace registration and installation — passed
- installed plugin reported version 0.9.1, enabled, one skill, and one MCP server
- installed-cache QA — 24/24 checks passed
- JavaScript syntax and Git whitespace checks — passed

Static and dry-run validation made no live AIDP or OCI calls.

## Contribution checklist

- tracking issue: #99
[...]

PR #103, opened 2026-08-31
fix(ask-aidp): re-apply Windows cmd.exe RCE fix dropped by the #100 squash

ahmedawan-oracle:
## ⚠ Security regression on `main` — re-applies the fix that PR #100 was supposed to land

PR #100 merged as `da57ba1`, but GitHub's PR-head cache was **desynced** after a force-push (the app/webhook connection was failing all session). The squash collapsed the PR's *cached* head — the original pre-fix commit `36c7437` — so the reviewed-and-verified security fix **never reached `main`**, and the Windows command-injection RCE shipped in the plugin.

Confirmed on current `main`: `normalizeSpawnTarget` still routes `.cmd`/`.bat` shims through `cmd.exe` with the broken `quoteWindowsCmdArg` caret quoting; `resolveCmdShimToNode` absent.

This PR re-applies **exactly** the reviewed fix (the two files are byte-identical to the verified commit `72b10b3` from PR #100's review), on top of current `main`:

- **RCE fix:** `.cmd`/`.bat` arguments are no longer routed through `cmd.exe`. A tool argument containing a `"` could otherwise terminate the quoted region and inject `&`/`|`/`>` commands (CVE-2024-27980 / "BatBadBut" class). Now resolves the shim to its Node script (`resolveCmdShimToNode`) and spawns `node <script> …args` with an argv array + `shell:false`, so every metacharacter is inert; fails closed if unresolved.
- **Error hardening:** pre-flight tool errors return `isError:true` results instead of JSON-RPC `-32000` protocol errors leaking internal stack traces.
- **`.gitignore`:** ignore `aidp.env` (the README setup copies a real-credential file there) and `node_modules/`.

### Verified on this branch
- RCE exploit probe (crafted `"`-bearing arg via a `.cmd` shim on real `cmd.exe`) **no longer executes** — sentinel not created; the arg reaches the CLI as one inert literal.
- `node --check` passes; `scripts/qa-claude.mjs` 23/23 in a clean env.
- Diff vs `main` is exactly the two files.

Please merge to close the live RCE. (If squash-merging, please confirm the PR head shows this branch's tip first — the #100 desync is why the fix was lost.)


mariomiola (commented):
Current npm-generated .cmd shims typically reference their JS entrypoint as %dp0%\..\package\bin.js. 
The fallback is capturing the literal %dp0%, then existsSync() fails and the CLI is refused. 
This may break the normal Windows aidp.cmd installation path the change is intended to secure.

ai/claude-code-plugins/ask-aidp/mcp/ask-aidp-server.mjs
```diff
@@ -1462,17 +1462,35 @@ function normalizeSpawnTarget(command, args) {
...
+}
+
+function resolveCmdShimToNode(cmdPath) {
+  // npm-generated .cmd/.bat shims invoke Node against a wrapped .js entry
+  // (e.g. `"%~dp0\node.exe" "%~dp0\node_modules\aidp-cli\dist\bin\aidp.js" %*`).
+  // Extract that .js path so we can spawn Node directly and bypass cmd.exe.
+  let text;
+  try { text = readFileSync(cmdPath, 'utf8'); } catch { return null; }
+  const dir = path.dirname(cmdPath);
+  const match = text.match(/%~dp0\\?([^\s"'\r\n]*?\.js)/i) || text.match(/([^\s"'\r\n]+\.js)/);
```
mariomiola: Please review the comments

ahmedawan-oracle: Good catch, @mariomiola — you're right, and it's fixed in `ff46fb1`.

The first cut only stripped `%~dp0`, so a modern npm shim referencing `"%dp0%\..\aidp-cli\dist\bin\aidp.js"` left the literal `%dp0%\..\` in the path, `existsSync()` failed, and the shim was refused — breaking exactly the normal Windows `aidp.cmd` path this change is meant to secure.

`resolveCmdShimToNode` now:
- scans **all** `.js` tokens in the shim (not just the first),
- strips the dp0 directory variable in every spelling — `%~dp0`, `%dp0%`, `%dp0` — plus its trailing separator,
- resolves the remainder against the shim's own directory with `path.resolve` so relative `..\` segments work,
- skips any token still carrying an unresolved `%VAR%`, and
- returns the first path that exists.

Verified on real shim **bytes** for three layouts — modern (`%dp0%` + `..\`), older (`%~dp0\node_modules\…`), and forward-slash — all resolve to the wrapped `aidp.js`. `cmd.exe` is still never invoked, so the RCE stays closed (exploit probe still fails to execute), and `node --check` passes.

mariomiola (approved):

PR #104, opened 2026-09-03
Update/adw iceberg sync

lucas-de-almeida-reis:
# Description

Three changes to the **ADW Iceberg External Table Sync** sample
(`data-engineering/adw-iceberg-external-table-sync/`): one bug fix in the notebook, two
configuration features, and the documentation that goes with them.

**1. Fix — `NameError: name '_ident' is not defined` on the apply step**

The identifier-validation helper was defined only inside the teardown cell, which ships wrapped
in a triple-quoted string and therefore never executes, while the engine cell calls it at two
live call sites. Any run with real work to do failed on every ADW in the fleet.

The dry run passed because both call sites sit on the apply branch, so the failure only appeared
after the plan already looked healthy. `_IDRE` and `_ident` moved to the helpers cell —
definitions only, always runs, ahead of both the teardown and engine cells. The teardown cell
keeps its own copy so it still stands alone when enabled.

**2. Feature — the mTLS wallet is now optional**

New fleet-wide flag `flags.use_wallet`, defaulting to `true` so existing deployments are
unaffected. With `false`, wallet extraction is skipped and the two `wallet_*` secrets are not
read at all: 4 credentials per ADW become 2, and the wallet bucket and its IAM policy are no
longer needed.

It is a config flag rather than something inferred from whether a wallet secret happens to
exist — a missing secret should surface as an error, not silently change how the job connects.
Each ADW then needs a TLS connection string in its `_dsn`; the connectivity probe at the end of
section 3 is what catches a DSN still set to the mTLS descriptor.

**3. Feature — `adw_user` accepts a mapping keyed by ADW prefix (optional)**

```yaml
adw_user:
  demo_adw1: SYNC_ADMIN
  demo_adw2: ADMIN

### CODE REVIEW

**Comment by ahmedawan-oracle** [comment]
## Deep review — all three claims verified by execution; one major: the fix resurrects an apply path where catalog-derived table names reach DROP DDL *before* `_ident` validates

@lucas-de-almeida-reis — reviewed by execution: the real notebook cells were run in order (base `07e4e9a` vs head `66cd3e4`) with recording fakes for `oracledb`/`oci`/secrets, driving the actual apply/teardown paths. All three PR claims hold, with executed proof:

### Verified ✔

**1. `_ident` fix** — reproduced on base: `_ident`/`_IDRE` exist only inside cell 10's triple-quoted string; the dry run genuinely passes (it returns before `admin_setup`), and apply dies with the NameError at the first live call site. On head: both call sites execute (`CREATE USER`/`GRANT`/ACL via `_ident(owner)`, `DROP TABLE`/`CREATE_EXTERNAL_TABLE` via `_ident(name)`), no NameError. The teardown's standalone copy is byte-identical in regex and AST-identical in logic (only docstrings differ) — no drift. Also swept for other members of this failure class: the set of names defined only in disabled cells has empty intersection with names referenced by live cells — `_ident` was the sole instance and the PR fully resolves it.

**2. `flags.use_wallet`** — with the flag absent or `true`, the secret-read set and `oracledb.connect` kwargs are byte-identical to base (verified by recorder diff). With `false`: exactly 2 secrets read per ADW, the `wallet_*` secrets are **never requested**, zero object-storage reads, wallet-free connect kwargs — the 12→8 / 4→2 claims are exactly right. The fail-loud contract holds both ways: `true` + missing wallet secret dies loudly in the config cell (no silent walletless fallback — your "deliberate flag, not inferred" design is real in the code), and `false` + mTLS descriptor DSN dies at the section-3 connectivity probe before discovery/apply, as documented. `ensure_wallets()`/`cleanup_wallets()` no-op walletless, so the disabled teardown/validate cells stay safe when re-enabled (executed).

**3. `adw_user` mapping** — single consumer chain (`cell 4 → adw["user"] → connect_admin`), so the mapping lands everywhere: executed the full apply, the unwrapped teardown, and the validate cell under `{demo_adw1: SYNC_ADMIN, demo_adw2: ADMIN}` — every statement ran on the right user's connection per ADW. Scalar form is byte-identical to base; the loud positional-list rejection with a copy-paste migration mapping is a nicely engineered back-compat break; unknown/case-mismatched keys raise with the offending keys named. An injection attempt through a mapping value never reaches SQL (only the `connect user=` kwarg) and doesn't influence secret names — verified against the full SQL log.

**Docs** — the README grant model was checked against the complete executed SQL inventory (19 admin-connection statements): every statement is covered by the documented set, the two documented failure rows map to real executed statements, and nothing executed falls outside the discussion. No leftover contradictions from the five docs iterations; the removed preflight script left no dangling references. Public-repo sweep: placeholders only, UPL/copyright header intact, diff touches exactly the four stated files. (The ADB-23ai-specific measurements — PDB_DBA carrying the two extra grants, DBA failing the DBMS_CLOUD WGO check — aren't reproducible here; they're internally consistent and plausible, taken as author-verified.)

### Major — statement ordering defeats `_ident` on the path this PR brings back to life

**Catalog-derived table names reach DROP DDL unvalidated.** The `_ident` validator itself is strong (verified against a hostile battery: quotes, spaces, semicolons, homoglyphs, embedded/trailing newlines — all rejected; the `\Z` anchor even beats the trailing-newline trick). But in `process_schema`, the default no-views create/recreate path interpolates the raw table name into `DROP VIEW <T>` / `DROP TABLE <RAW>` literals that execute **before** `_create_ext` calls `_ident`, and the drop branch never calls `_ident` at all. Executed end-to-end on head: a catalog answering `SHOW TABLES` with `EVIL'; DROP TABLE ADMIN.EXT_REGISTRY_V4; --` flows through discovery into the plan, and apply executes

```
BEGIN EXECUTE IMMEDIATE 'DROP VIEW EVIL'; DROP TABLE ADMIN.EXT_REGISTRY_V4; --'; ... END;
```

— the payload breaks out of the `EXECUTE IMMEDIATE` literal and runs as the schema user; `_ident` raises only afterward, too late. The interpolation text is pre-existing, but on base this entire path was dead (the NameError fired first), so this PR is what makes the sinks reachable — while the cell's own comment says the name is validated "before it reaches the DROP literal". Fix is small: validate `name`/`raw` (both `_ident`) at the top of `process_schema` — or at minimum before the first DROP in each branch, including the drop branch. Vector realism: a hostile (or just quote-containing) table name in a shared/source-controlled Iceberg catalog is exactly the input this sample syncs.

### Minor

- **`flags.use_wallet: "false"` (quoted) is truthy** — `bool(C(...))` on the string yields `True`, so an operator who followed the walletless walkthrough (which deletes the wallet secrets) gets a misleading credential-404 about `<prefix>_wallet_zip` instead of a hint about the quoting mistake. One `isinstance(bool)` check with a loud message closes it; the idiom is pre-existing on the other flags, but this is the only one whose misparse changes which secrets must exist.
- **A null mapping value becomes the literal DB user `None`** — `adw_user: {demo_adw1: }` (blank while editing) resolves to `str(None)` → connects as user `None` and dies as a password error; inconsistent with the absent-key case (falls back to ADMIN, documented). Same `str()` coercion turns YAML-1.1 `yes` into user `True`. Reject non-string/None mapping values with the same loud style as the unknown-key check.
- **The sample YAML's own `adw_user` example doesn't run against the sample itself**: the commented example uses `pb_adw1`/`pb_adw2` while the file's `adw_prefixes` are `demo_adw1`/`demo_adw2` — pasting the file's own example raises the unknown-key RuntimeError (executed). Looks like residue from a private environment (`pb_`, not sensitive); the notebook's cell-4 comment already uses the right keys — align the YAML.

### Notes

- The walletless TLS/mTLS-descriptor mismatch surfaces as the raw driver error (e.g. DPY-6005); the actionable remedy is in the stdout banner just above but not attached to the exception — wrapping the connectivity-probe failure with the walletless hint (mirroring `_vread`'s style) would make it self-contained in scheduled-job logs.
- `_ident(None)` returns the literal identifier `'NONE'` (`str(x).upper()` before matching) and there's no length cap (>128 chars passes, fails later as ORA-00972). Not injectable; worth a two-line tighten while you're in that function. Pre-existing.
- The "Adding a new ADW" walkthrough still says "Four secrets … Upload the wallet" unconditionally — Step 1 and Step 3 got walletless skip-notes in this PR, so this section missing one is an inconsistency of the PR's own (good) convention.

**Bottom line:** the engineering here is careful — the fix is real and complete for its class, both features do exactly what the PR text says (verified to the level of secret-read sets and per-connection SQL logs), the docs match the code unusually well, and the loud validation style is consistent. The one thing I'd hold for is the DROP-ordering major: it's a small hoist of the existing validator, but as shipped the PR re-arms a path where a quote in a table name executes DDL. The three minors are quick config-validation fixes. Happy to re-verify on the next push. 🤖 Deep review by Claude (Fable)

lucas-de-almeida-reis:
Thanks — thorough review, and it caught a real one. All four items addressed in `4a73c0c`.

## Major — confirmed and fixed

You're right about the mechanism, and I reproduced it before changing anything. Built a
recording-cursor harness around the real `process_schema` extracted from the notebook:

| | statements executed | carrying the payload |
|---|---|---|
| base `66cd3e4` | 3 | **3** |
| head `4a73c0c` | 0 | **0** |

On base the SQL came out exactly as you described:

    BEGIN EXECUTE IMMEDIATE 'DROP VIEW EVIL'; DROP TABLE ADMIN.EXT_REGISTRY_V4; --'; ... END;

Fixed by hoisting the validator to the top of each branch, including the drop branch, as
you suggested. Two things beyond your note:

- `_ident(T + RAW_SUFFIX)` on the view-mode drop as well — `RAW_SUFFIX` comes from the
  config rather than the catalog, but it reaches the same interpolation.
- Caught `ValueError` alongside `DatabaseError`. This turned out to matter as much as the
  hoist: `_ident` raises `ValueError`, which `except oracledb.DatabaseError` did not
  catch, so the exception escaped the worker thread and **aborted the entire schema
  slice** — every remaining table in that slice went unprocessed. It's now an error on
  the one table, and the slice continues.

That second part also fixes a class of ordinary names, all legal upstream, that on base
took down a whole schema slice after the DROPs had already run and been swallowed:

| name | why it is not a legal Oracle identifier |
|---|---|
| `my-table` | hyphen |
| `order.items` | dot |
| `2024_sales` | leading digit |
| `tbl name` | space |
| 130 characters | over the 128 limit |

Each of these now reports cleanly as `invalid Oracle identifier: 'MY-TABLE'` against that
table alone.

## Minors — all three fixed

1. **Quoted flags.** Added a `CB()` reader that requires a real YAML boolean. Applied to
   **all six** flags rather than just `use_wallet` — they all misparse identically, and
   `create_views: "false"` silently creating views is no better. Rejects `"false"`, `"no"`
   and `0` with a message that points at the quoting.
2. **Mapping values.** Rejects `None`, booleans (YAML 1.1 `yes`) and blank strings, naming
   the offending keys, in the same loud style as the unknown-key check. Valid values are
   `.strip()`ed, so `"  SYNC_ADMIN  "` works.
3. **YAML example prefixes.** Aligned to `demo_adw1` / `demo_adw2`. Good catch.

## Notes — all three done

- The connectivity probe now attaches the remedy to the exception, and the message differs
  by `USE_WALLET`, so it also covers the inverse mistake (wallet on, TLS descriptor DSN).
- `_ident` rejects `None` and non-strings, and caps length at 128.
- "Adding a new ADW" now reads "Two secrets, two credentials" with the walletless notes on
  steps 2 through 4, matching the convention Step 1 and Step 3 already had.

I left the pre-existing `# SECURITY:` comments on the `admin_setup` and `_create_ext` call
sites as they were.

Ready for re-verification whenever you are.

ahmedawan-oracle:
## Re-verification — every claimed fix confirmed by execution; three small residuals found and closed in `2deac88`; approving and merging

@lucas-de-almeida-reis — verified `4a73c0c` with the same executed methodology (each finding reproduced on `66cd3e4` first):

**All confirmed closed:** the end-to-end injection now executes **zero** payload-carrying statements in all four branch scenarios (create/drop × view modes) with per-table errors and the slice continuing — your reproduction table matches mine exactly; the widened `except` genuinely fixes the slice-abort class (and doesn't over-swallow — a `DatabaseError` on CREATE still surfaces per-table identically); `CB()` rejects quoted booleans on all six flags **before any secret read**, with back-compat byte-identical for real booleans; the mapping validation rejects null/`yes`/blank values loudly with the remedy in the message; the YAML example runs against its own prefixes; the connectivity probe wraps the driver error with a `USE_WALLET`-conditional hint; `_ident` gained the None/length guards; the walkthrough got the walletless notes. One precision correction for the record: on the old head the >128-char case failed per-table as ORA-00972 (`DatabaseError`), not via the escaping `ValueError` — your other four rows are exactly right.

**Three residuals surfaced during re-verification — fixed in `2deac88` (pushed to this branch, verified by re-running the full battery):**

1. **Length-cap regression (introduced by `4a73c0c`)**: `RAW = _ident(T + RAW_SUFFIX)` rejected legal 124–128-char tables even in default no-view mode (where RAW only feeds the swallowed leftover-DROP and grant capture), and in view-mode drop the error fired **between** the two drops — view gone, registry row kept, failing drop recurring every run. Fix: `_raw_name(t, live)` — charset always checked (the suffix reaches the same literals), the 128 cap only when the raw table must exist as a live identifier; the drop branch validates before `DROP VIEW`. Executed: 123/124/128-char names sync again; view-mode drop issues both drops; injection battery still all-zero.
2. **The teardown cell's standalone `_ident` had drifted** — it kept the old body (accepts `None`→`NONE` and >128 names; injection regex was identical). Now carries the same guards as the helpers-cell copy.
3. **Scalar `adw_user` was still unvalidated** (pre-existing — your fix covered the mapping shape): `adw_user: yes` connected the whole fleet as user `"True"`, the exact failure your own new error message warns about. Now rejected in the same loud style; valid scalars are `.strip()`ed like mapping values.

Cosmetic leftover you can fold into any future touch: `CB()`'s rejection text for int values (`use_wallet: 0`) says "a quoted value is read as text" — the int case deserves its own wording.

Approving and merging — the sample is in genuinely better shape than when this PR started: the fix round closed everything, added guards beyond the asks, and survived a fresh adversarial pass with only the three items above. Thanks for the fast, rigorous turnaround. 🤖 Deep review by Claude (Fable)

ahmedawan-oracle (approved):
Verified by execution across two review rounds — injection closed end-to-end, all findings addressed. See review comments.

PR #107, opened 2026-09-10
Features/sample code fixes

navid-heydari:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

Please delete options that are not relevant.

# How Has This Been Tested?
 Tested in AIDP workbench.

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

# Checklist:

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes


lucas-de-almeida-reis (commented):
There's real value in here — the SDK generation finding (`datalake.{region}` + `/20260430`, and that
all 18 clients moved) and the clone tests in T4 are both worth having. But the PR bundles four
unrelated changes across 12 files, which makes it slow to review and means the good parts can't land
until everything else is resolved.

**Recommendation: close this PR and reopen as four focused ones.** Details below.

## How this repo works

Worth stating since it drives most of the feedback:

- **`README.md` at the root is the catalog, and it's the only entry point.** A sample that isn't
  listed there is effectively invisible — nobody browses directories. `CONTRIBUTING.md` requires the
  row, and specifies the format and which section it goes in. Neither new sample here is listed.
- **One PR = one concern.** Either a single new sample, or a fix to one thing. Not both, and not
  several samples at once. It keeps review scoped and lets unrelated work land independently.
- Samples are **written for an external reader** who has their own tenancy and no context from us.
  That's the bar for anything in the catalog.

## Problems to fix on the way out

**Applies to every split PR:**

- Add the row to the root `README.md` catalog.
- Remove internal environment identifiers: `navid-dev-sandbox (us-ashburn-1)` in three READMEs, a
  live AIDP OCID hardcoded as a default in `test_aidp_lineage_api.py:41`, and the real Object Storage
  namespace and bucket in the committed outputs of `Verify_Data_Lineage.ipynb`. Read the OCID from the
  environment with no default, and clear notebook outputs. As written, an external reader's run
  targets someone else's resource.
- Drop the whitespace-only changes: `.gitignore` (adds a blank line, removes the trailing newline) and
  `transformation/merge/requirements.txt` (`-pytest`/`+pytest`).
- Fill in the PR description. Every checkbox is currently ticked, including "I have added tests that
  prove my fix is effective", while the Description is unedited template text.

## The four PRs

**1. Fix — type mismatch in `telecommunications_medallion_architecture_demo.ipynb`.** The fix is real
and welcome. Rewrite the added cell ("## You Got an Error during Insertion! / RCA is the mismatching
type") as a short explanation of the mismatch and how to avoid it. This one can merge quickly.

**2. Fix — `slowly_changing_dimension_template.ipynb`.** Move `jinja2` to `requirements.txt` instead
of `%pip install` in the notebook, and drop the committed pip install transcript.

**3. New sample — `data-engineering/lineage/`.** Keep it; the SDK generation finding is the reason.
Two things to resolve first:
- The README and `TAKEAWAYS.md` contradict each other. `TAKEAWAYS.md` says the sibling's 404-based
  reasoning proves nothing and ends with "Worth updating that README" — and the README is unchanged.
  Reconcile them, and don't ship the review note as published text.
- The headline says "Proves AIDP's own lineage API is released and live", but "Known gap" documents
  that `fetchLineage` rejects every `anchorNode` with `400`. State that up front, so a reader knows
  before setting up a profile that they won't get a graph out of it.
- `pytest.ini` and the suite need a personal `~/.oci/config`, so they aren't runnable as a demo — one
  README line saying it's a maintainer harness would help.

**4. New sample — reshape `capability-probes` as a clone + metadata sample.** The comparative framing
isn't needed here and isn't ours to publish: competitive positioning against Snowflake and Databricks
is produced by the product team, and only they decide whether it's released. So drop Part A (the
`oci raw-request` endpoint enumeration) and T1/T2, which exist only to show that unsupported syntax
fails.

What remains is a good sample under a different title. T4 is the strongest thing in this PR: shallow
vs deep clone, `DESCRIBE DETAIL`/`HISTORY` proving the zero-copy property, copy-on-write divergence
with explicit expected counts, and the `VACUUM`-breaks-the-clone warning — Delta behaviour worth
documenting on its own terms. T5 complements it; T6/T8 cover attaching and reading metadata in SQL.
Suggest **"Zero-copy clone and table metadata on AIDP Delta"**. The `run()` harness can go — what's
left works, so failures should fail loudly.

Please drop T9: it applies one mask shape (`concat('XXX-XX-', right(c,4))`) to every tagged column
regardless of type — on an email that gives `XXX-XX-.com` — and interpolates a column name straight
from the registry table into generated DDL. A column-masking sample is worth writing, as its own piece.

Happy to review each one as it goes up — PRs 1 and 2 should be quick.

cc @ahmedawan-oracle @navid-heydari

ahmedawan-oracle:
## Deep review of `d589819` — solid investigative work on lineage and a correct diagnosis of the medallion failure, but the PR ships real tenancy identifiers, the medallion "fix" leaves the failing cells in place, the SCD template misses NULL transitions, and the two new docs contradict each other on the very method one of them criticises

@navid-heydari — thank you for digging into lineage properly; the split between "API released" and "graph populated", the A2/A6 control tests and the xfail tripwire are good methodology, and the medallion root cause is right. Method for this review: everything below was established by execution where this host allows it — the shipped `test_aidp_lineage_api.py` collection and no-config behaviour, the notebook helpers and every probe statement through the Spark 3.5.9 + Delta 3.2.0 parser, the lineage extractor on 30 query shapes, the SCD templates rendered and emulated on in-memory data, the medallion generator's inferred schema run through Delta's own `SchemaMergingUtils.mergeSchemas`, and the public `aidataplatform-sdk`/docs for the REST claims. Catalog table writes cannot run on this Windows host and no live AIDP call was made, so write-dependent behaviour is inferred from parser + Delta 3.2 source and labelled. Every finding survived independent adversarial re-verification.

### Verified working ✔

- **Lineage notebook**: the extractor is correct for the two demo hop shapes — `references()`-based resolution through aliases and aggregates, `COALESCE`/`CASE` across two sources, `SELECT *` expansion, self-join de-duplication, WHERE-only columns excluded exactly as documented, non-catalog leaves flagged with a placeholder so check 6 fails as designed; every py4j accessor it relies on exists on Spark 3.5.9; 11 `check()` sites expand to exactly the 19 checks claimed; clean-up drops all five tables and the schema.
- **Lineage test module**: 17 tests collect as 13 pass-designed + 4 xfail, matching the README; the anchorNode matrix, env-var names, run commands and `requirements.txt` all match the code; the REST shape you document (`POST /20260430/aiDataPlatforms/{id}/actions/fetchLineage|exportLineage`, `datalake.{region}` host template, `datahub-dp` prefix, "(Preview)") is confirmed in the public SDK.
- **Capability probe**: the Spark 3.5 parser rejects `SET MASK`, `SET TAGS`, `CREATE MASKING POLICY` and `SET ROW FILTER` as the README expects; `SHALLOW CLONE` parses; the redacting-view workaround (T3) executes correctly; the harness catches eager DDL failures without aborting.
- **Medallion**: your RCA direction is right — the generator infers `double`/`bigint` where the bronze DDL declares `DECIMAL(10,3)`/`DECIMAL(8,2)`/`INT`, and Delta 3.2's schema merge throws `DELTA_FAILED_TO_MERGE_FIELDS` on the original write (Delta source confirms V1 `saveAsTable(overwrite)` keeps that check); the cast cell and the explicit silver select list both produce schemas identical to the DDL; the `float()` wrap in the ML cell is a necessary fix (`Decimal * float` raises).
- **SCD**: the Jinja2 templates render the committed SQL byte-for-byte, `MERGE INTO` parses under the Delta parser, an in-memory emulation reproduces the committed result, a second run with the same staging is idempotent and an unchanged row produces no new version; removing the `SparkSession.builder` boilerplate and using `%pip` are both right for AIDP kernels.
- **Hygiene**: no credentials, keys, IPs or personal e-mails anywhere; the `.gitignore`/`requirements.txt` deltas are harmless whitespace.

### Majors — identifiers and content that must not ship in a public repo

1. **A real DataLake OCID is the default in `test_aidp_lineage_api.py:41`** (type: `ocid1.aidataplatform…`, feeding `DP_BASE`, `LEGACY_BASE` and the anchor matrix). Anyone who follows the README's `pytest test_aidp_lineage_api.py -v` without exporting `AIDP_DATALAKE` signs every request against your resource and `A0` fails with `404 NotAuthorizedOrNotFound` — the exact ambiguous error your own TAKEAWAYS warns about — so the "proof of release" reads as a failure for everyone but you. Make the variable required (skip/exit with a clear message when unset), document it as a prerequisite, and rewrite history before merge — the value is in commit `03c4cba`.
2. **`Verify_Data_Lineage.ipynb` cell 18 commits three `oci://<managed bucket>@<namespace>/managed/…` URIs** in its outputs — the Object Storage namespace is a tenancy-wide identifier and the bucket is the DataLake's internal managed bucket. The check only needs "files live under the location"; print the path after the authority, or clear that cell's outputs.
3. **The author's sandbox tenancy name appears in three files** (`TAKEAWAYS.md:3`, `capability-probes/README.md:66`, probe notebook cell 0). Replace with "a development tenancy in us-ashburn-1".

### Majors — the medallion "fix" does not fix the notebook

- **Cells 5 and 12 are byte-identical to `main`, outputs included**, so `Run All` still halts at cell 5 for every user — the notebook's own committed AIDP metadata records `result_type: error` for cells 5, 12 and 27 from your 2026-07-31 run, while the same cells display `main`'s "Successfully ingested 599191 raw records" / "Successfully processed 599179 cleaned records" directly above the new markdown "You Got an Error during Insertion!" and "again the same error above". The workaround cells 7 and 14 ship with no execution count and no output, and the gold/ML cells carry counts 38/41/42/54 with cleared outputs. Cell 22's output prints doubles (`1.297370033E7`, `369.98`) that the PR's decimal code path cannot produce. It is a spliced artifact, not a run record.
- **`.option("overwriteSchema","true")` on the three gold writes silently replaces the DECIMAL/INT schema declared in cells 17 and 19 and drops `created_at` from all three tables.** Executed: the gold SQL over a silver typed per its DDL yields 8 / 3 / 5 type mismatches (`decimal(21,3)` vs `DECIMAL(10,3)`, `double` vs `DECIMAL(5,2)`, `bigint` vs `INT`, `decimal(28,2)` vs `DECIMAL(10,2)` …); with `overwriteSchema` the DataFrame's types win and the DDL cells become dead code. The note "you already fixed the Silver and Bronze and learned the lesson, you can just push for gold and wrap this up" teaches the opposite of the lesson. Either `CAST` each aggregate to the declared type and add `current_timestamp() AS created_at`, or drop the DDL cells; not both.
- **The RCA text misattributes the causes**: a cast-only bronze DataFrame *without* `ingestion_timestamp` merges identically to the DDL (missing columns are filled with NULL), so "add the missing column" is not part of the bronze fix; silver fails because the DataFrame carries an *extra* column (`ingestion_timestamp`) the table lacks, not the omitted `processed_timestamp` that cell 14's comment blames.
- Fix in place: build `df_bronze` with an explicit schema (or the casts) in the generating cell, select exactly the DDL columns in silver, delete cells 6/7/13/14 and the informal markdown, run top-to-bottom on a clean schema and commit either one consistent set of outputs or none.
- **For the maintainers (pre-existing, outside this PR's diff):** the same generator-vs-DDL mismatch breaks 6 of the other 11 medallion demos on `main` (energy, financial_services, manufacturing, real_estate, retail; transportation additionally raises `NameError: ROUTE_TYPES` before reaching it — insurance is fine because it casts). Executed through the same Delta merge check. The catalog advertises 12 working demos; this PR touches one. A shared fix pattern in a follow-up would be worth more than this notebook alone.

### Majors — SCD template

- **Change detection is NULL-unsafe.** Both templates render `target.col != source.col`; in Spark `NULL != 'x'` is NULL, falsy in `WHEN MATCHED AND` and in `WHERE`. Executed emulation: target `email` NULL → source `alice@example.com` produces zero updates and zero inserts (change lost, Alice stays current with NULL email); value → NULL likewise. Render `NOT (target.{{col}} <=> source.{{col}})` or `IS DISTINCT FROM`, and add a NULL-transition row to the staging data so the demo shows it. Pre-existing in the template, but this PR re-runs and re-ships it as the reference SCD2 sample.
- **The target moved to a hard-coded `lake.scd` that nothing creates** — cell 2 drops `lake.scd.dim_customer` and cell 3 runs `CREATE SCHEMA IF NOT EXISTS lake.scd`, but no `CREATE CATALOG` exists, cell 0's prerequisites (unchanged from `main`) mention only jinja2, and the root README row is unchanged. Sibling notebooks either create their domain catalog (`CREATE CATALOG IF NOT EXISTS` in 38 files) or use `default`, as this PR's own lineage notebook does. On a DataLake without a catalog literally named `lake`, cells 2/3/12 fail at analysis time. Parameterise `CATALOG`/`SCHEMA` in a first cell, move the DROP after the CREATE, and state the requirement.
- Minor: the committed cell-11 output ("Executing Update SQL: …") is produced by two `print()` lines that ship **commented out** in cell 5 — the committed code cannot produce the committed output (introduced by this PR's re-run); cell 9 still labels Bob's row `# same as before` while his status changes and the output shows him versioned, so the no-op path is never demonstrated; duplicate keys in staging match one target row twice (Delta MERGE fails; the insert step would emit two current rows) and the notebook never states the uniqueness precondition; the `%pip` cell duplicates cell 0's "install on cluster via requirements.txt" instruction.

### Majors — lineage extractor accuracy is overclaimed

Cell 23, the README and the TAKEAWAYS scoreboard say lineage "can be extracted … accurately at both table and column granularity — proven 19/19", with WHERE/JOIN-ON columns as the only stated limitation. Executed with the verbatim cell-4 code on Spark 3.5.9: **silent wrong or empty column lineage** (printed as `(literal)` by cell 14's formatter) for a window function, a `Project` over an `Aggregate` (SQL subquery or DataFrame `.agg().select()`), `LATERAL VIEW explode`, and the second branch of a `UNION`; the whole column map is **empty with no warning** when the top node is `Sort`/`GlobalLimit`/`Distinct`/`Filter` (HAVING)/`WithCTE`/`Union`/`Except`; a CTE adds a spurious `<CTERelationRef>` leaf so check 6 would **fail falsely**; scalar/IN subquery sources are **missing** from table-level lineage. The 19/19 holds only because the demo pipeline is two `Project`/`Aggregate` roots. Resolve attributes transitively (walk `Alias` → child references through Project/Aggregate/Window/Union, include `subqueriesAll()`, resolve `CTERelationRef` through `cteDefs`) and warn when the root is not Project/Aggregate — or narrow the claim to what is proven. Note also that `_output_exprs` "works" only through its exception fallback: py4j `hasattr()` is always true, so the guard never filters.

### Majors — capability probe and the Part A contradiction

- **Five probe statements are malformed for Spark 3.5, so supported features are recorded UNSUPPORTED**: T4c/T4d `SELECT … FROM (DESCRIBE DETAIL|HISTORY t)` is a `PARSE_SYNTAX_ERROR` (top-level `DESCRIBE DETAIL/HISTORY` parse fine — post-process in Python); T5c's `FROM (DESCRIBE t)` likewise; T5a `CREATE TABLE a.b.c LIKE a.b.d` — Spark 3.5's grammar accepts only 2-part names for `LIKE` (the 2-part form parses); T6a `'owner'` is a reserved table property. T1a uses the SQL-bodied `CREATE FUNCTION … RETURN` syntax that arrives in **Spark 4.0** (SPARK-46057), so it fails on 3.5 regardless of masking and T1b/T2c then reference a function that never exists. `DEEP CLONE` is rejected by the OSS Delta 3.2 grammar (docs: shallow only) — the Oracle `3.2.0-oci` build may differ, and the probe should say so. These are Spark-core grammar facts, identical on AIDP's 3.5.0.
- **The `run()` harness hides why**: for every `ParseException`, PySpark's message starts with a newline, so `str(e).splitlines()[0]` prints an empty reason after the arrow; Py4J errors print only `An error occurred while calling oNN.sql.`; and `spark.sql()` without an action masks runtime failures when `show=False` (`raise_error(...)` → SUPPORTED). Use `str(e).strip()`, `e.java_exception.getMessage()`, and always execute. Cell 23 (T9) also calls `spark.table()`/`.collect()` outside `run()`, so a failed setup aborts `Run All` before the summary and the cleanup — the opposite of cell 0's promise.
- **README vs notebook**: the README says "Part B SQL tests T1–T5" while the table and notebook define T1–T9; cell 0 says "Verified live" but all 27 cells have no execution count and no output, and "Findings (recorded)" records no per-test result — only the control-plane 404 list.
- **Internal contradiction between the two new docs.** `TAKEAWAYS.md` §3 (lines 41-69, scoreboard line 157) states that on this surface a nonexistent route and an unpermitted route return byte-identical 404s, that the "`/catalogs` → 200 so this 404 means absent" inference is wrong, that the old `aidp.<region>/20240831` generation is "a full generation stale", and that the sibling capability README is "worth updating … to cite SDK absence rather than HTTP 404s". The same PR ships `capability-probes/README.md:65-69` and notebook cell 1 with exactly that inference, on exactly that generation. Rewrite Part A per your own rule (new host/version, a bogus control path, absence argued from the SDK) and drop the "worth updating" homework and the "this repo has made before" framing for a file this PR introduces.

### Minors — docs, tests, process

- **Service naming is stale against the current SDK**: README/TAKEAWAYS/test docstring/notebook call the service `SemanticCatalog` and list "18 clients … SemanticCatalog … no AsyncOperations". The SDK shipped that name in v4.1.0 (2026-08-07), but current `main`/v4.2.1 exposes `DataLineage` (`DataLineageClient`, CLI group `data-lineage`, operations `fetchEntityLineage`/`exportLineage`) plus `AsyncOperations`, and its CLI reference still documents `aidp.<region>` as the default endpoint — so "the whole SDK moved" is overstated. Pin the observation to a date/SDK version and use the current names.
- **Field-sales voice in a public sample**: "Two things to carry into a customer conversation", "Positioning vs Databricks … GA … don't oversell it", "The one question for Oracle" — no sibling README reads like this, and TAKEAWAYS publicly asserts an unconfirmed product gap from one tenancy on 2026-08-16 while the README says "Start here". Keep a neutral "Known limitation observed on <date>" and move the positioning notes internal.
- **Test module robustness**: with no `~/.oci/config` the session fixture raises `ConfigFileNotFound` — 13 ERRORs (no skip) and the four Part-B tests report "xfailed" without a single request being sent, so a user without credentials sees the "known gap" outcome for the wrong reason (`OCI_CONFIG_FILE` is only honoured when `~/.oci/config` is absent). Exact-message assertions (`== 'Invalid anchorNode'`, `code == 'InvalidParameter'`) will flip the "proof" red on any wording change in a Preview API; A7 depends on the legacy gateway staying up (which TAKEAWAYS calls stale); fixed 60 s timeout, no retry; the docstring omits `AIDP_DATALAKE`; `pytestmark = []` is a no-op; `pytest.ini` is the first in the repo (siblings use `conftest.py`).
- **Repo policy**: none of the six commits carries `Signed-off-by` (CONTRIBUTING requires it); no issue reference; the PR template placeholders are still in the body while every checklist box is ticked, including "added tests that prove my fix" (the medallion/SCD changes have none; the lineage tests need a live tenancy); the root README Sample Catalog and Repository Structure tree were not updated for the two new folders (CONTRIBUTING says "must", and `data-engineering/lineage/` is a new top-level folder outside its placement list — discuss the section in the issue).
- **Commit hygiene**: `d589819` is an empty commit; `8a91ffc` deletes seven `.gitignore` lines that `69e9316` re-adds (net: one blank line and a lost trailing newline); `083b8f8` downgrades `pyspark` to 3.5.0 and `cb51768` "Upgrade pyspark to version 3.5.2" restores it (net: a trailing newline). Drop both hunks and squash to two or three signed commits.

**Bottom line:** keep the lineage investigation and the probe idea — they are useful — but before this can merge: scrub the OCID, namespace/bucket and tenancy name (with a history rewrite for the OCID), repair the medallion notebook in place and drop `overwriteSchema`, make the SCD comparison NULL-safe and parameterise the catalog, narrow the extractor's accuracy claim or fix the plan walk, fix the five malformed probe statements and the harness, reconcile Part A with your own TAKEAWAYS §3, update the SDK naming, and land the process items (sign-off, issue, template, catalog rows). Happy to re-verify. 🤖 Deep review by Claude (Fable)

navid-heydari:
@ahmedawan-oracle  -- Lucas Suggested the splitting this into 4 PRs, since this is just a Sample and adjustments on my findings, do you want that or fix in place? either way works for me but wanna confirm from your side as well.

ahmedawan-oracle:
@navid-heydari — answering the open question here: fix-in-place with four single-concern commits works for me as a reviewer (Lucas asked for four PRs; that remains the maintainers' call), and #116 is the tree that now matters — the deep re-review is posted there: https://github.com/oracle-samples/oracle-aidp-samples/pull/116#issuecomment-5682502819. Since #116 replaces this PR and this branch still carries the OCID in its history (`03c4cba`), please close #107 as superseded. 🤖 Deep review by Claude (Fable)

navid-heydari:
Closed this one as per the discussion above and continuing the remaining items on #116

PR #108, opened 2026-09-11
Add AWS → Oracle AIDP Migrator (Claude Code plugin)

mppotnuru:
New Claude Code plugin under `ai/claude-code-plugins/`: a migration assistant that inventories an AWS data stack and translates Athena SQL and Glue ETL to Spark on AIDP via four verbs — inventory / plan / migrate / verify. It emits reviewable artifacts and flags anything it cannot convert safely; **applying the output to AIDP is manual today** (nothing is written to an AIDP workspace).

- Athena→Spark SQL via **sqlglot** (AST transpiler, not regex); Glue ETL→PySpark; S3→OCI via a generated **rclone** job. Glue Data Catalog is inventoried and planned (no DDL emitted yet); EMR/SageMaker translators are stubs.
- Deterministic-first: anything unsafe is **flagged**, never guessed. Visual `report.html` shows before/after + flags.
- Verified offline: 369-test stress suite (translated output executed on Spark 3.5) + unit suites — 377 passed / 548 subtests under pytest.
- MIT licensed. A Codex-packaged variant is proposed separately in #109.

*(Description refreshed at merge time to match the repositioned README — see review threads for the three validation rounds.)*


ahmedawan-oracle:
## Deep review — the offline pipeline, tests, fixture and secret hygiene are solid; but the plugin cannot be installed or run as shipped, both translators silently change semantics in routine cases, and the "verified live end-to-end" claim is not backed by migrator output

@mppotnuru — thank you for a substantial contribution; the AWS half of the migration story is genuinely needed and the deterministic-first design is the right instinct. This review was done by execution, not by reading: the 19 tests, the four-verb fixture pipeline, `claude plugin validate`, a wheel build + non-editable install into a fresh venv, the MCP server driven over stdio, a 113-case Presto battery whose Spark output was **executed on PySpark 3.5.9** (the AIDP kernel's Spark minor), 25 hostile Glue jobs, botocore-Stubber runs of the inventory, a recording fake for the live seed/teardown scripts, and every binary in `docs/` (pptx/pdf text layers, all 31 video + 9 gif frames viewed). The `aws_aidp/`, `tests/`, `scripts/`, `fixtures/`, README, TESTING and pyproject trees are byte-identical to #109, so every engine finding below applies to both PRs; #109 gets its own comment for the Codex-specific items.

### Verified working ✔

- **Tests and offline pipeline**: `pytest tests` → 19 passed (8 Athena + 8 Glue + 3 S3, matching the body); `inventory --fixture demo → plan → migrate --demo → verify` runs clean from a neutral cwd (plan 113 assets; migrate ok=20 needs_review=13 planned=80 error=0; verify PASS 20 / REVIEW 13 / SKIP 80 / FAIL 0); 18 of the 20 translated Athena queries analyse and execute on Spark 3.5.9.
- **Honest flagging where it exists**: unparseable SQL is flagged `parse_failed` and returned byte-identical; `histogram()` and `UNNEST … WITH ORDINALITY` are flagged as documented (POSEXPLODE alias order verified on Spark); `ApplyMapping`/`ResolveChoice`/… `.apply` forms are flagged and left in place; the Glue translator is idempotent on all 8 fixture jobs and all 25 hostile cases; receiver aliasing (`ctx`/`glue_context`/`glueContext`), multi-line balanced-paren argument lifting and the `s3://`→`oci://` rewrite (f-strings, `%`-templates, `s3a://`) work.
- **Secret hygiene**: no OCIDs, keys, e-mails, hostnames or account ids anywhere (68 regex hits reviewed, all placeholders/synthetic); the fixture is fully synthetic (account 999900001111, 0 PII-like literals) and its builder is deterministic apart from a timestamp; pptx/pdf metadata clean; the mp4/gif are rendered cards, not screen recordings — no console, tenancy or path appears in any frame; `.gitignore` covers `demo.env`, `.aidp/`, `.env`.
- **Manifest validity**: `claude plugin validate` passes (one warning: `category` belongs in marketplace.json); command frontmatter and `mcpServers: "./.mcp.json"` are documented forms; the SKILL description is action-oriented; the generated rclone job uses `rclone copy` (never sync/move) with real flags and a temp config, so nothing moves until the user runs it.
- **HTML report escapes user text** (`<script>`/`` in asset names and SQL come out as entities); CLI exit codes are consistent (migrate 1 on error, verify 1 on FAIL / 2 on missing input).

### Blockers

1. **The MCP server is dead on a fresh install.** `pyproject` declares `mcp>=1.2` with no upper bound; today that resolves to **mcp 2.2.0**, where `mcp.server.fastmcp` no longer exists. `aws_aidp/mcp_server.py:18` fails to import and the `except ImportError` rewrites it into `The MCP SDK is required. Install it with: pip install -e '.[mcp]'` — i.e. it tells the user to run the command they just ran. Executed: `pip install -e '.[mcp]'` → mcp 2.2.0 → `aws-aidp-mcp` and `python -m aws_aidp.mcp_server` both exit 1. The server works only with mcp 1.x (verified: handshake + 4 tools on 1.26/1.30). Fix: `mcp>=1.2,<2` (or port to `mcp.server.mcpserver.MCPServer`), surface the real ImportError, add a smoke test that imports the module and lists tools.
2. **Nothing installs the CLI for a plugin user, so every command and the MCP server fail after `/plugin install`.** `.mcp.json` spawns a bare `aws-aidp-mcp` (exit 127 on a clean PATH; Claude Code installs Node deps only, never pip); the plugin never references `${CLAUDE_PLUGIN_ROOT}` (0 hits); `SKILL.md:16` says `pip install -e .` (assumes the cwd is the repo — a marketplace plugin lives under `~/.claude/plugins/cache/…`) "or `pip install aws-aidp-migrator`" — **that package does not exist on PyPI** (404). Counter-proof that the fix is viable: `PYTHONPATH=<plugin root> python -m aws_aidp.mcp_server` handshakes and lists the four tools. Fix: `.mcp.json` → `{"command":"python","args":["-m","aws_aidp.mcp_server"],"env":{"PYTHONPATH":"${CLAUDE_PLUGIN_ROOT}"}}`, document a one-time `pip install "${CLAUDE_PLUGIN_ROOT}[mcp]"` (or a SessionStart hook like the databricks-migrator), and have the commands fall back to `python -m aws_aidp.cli`. Related: `fixtures/` is not packaged (`packages.find include=["aws_aidp*"]`, path resolved as a sibling of the package) — a wheel/non-editable install breaks `--fixture demo` and the MCP `inventory(fixture="demo")` tool (executed: exit 2 `fixture not found: …site-packages\fixtures\…`).
3. **The plugin still carries its personal-repo identity, and that repo is private.** `README.md:47-48` installs from `/plugin marketplace add mppotnuru/aws-aidp-migrator` + `/plugin install aws-aidp-migrator@aws-aidp-migrator` — `gh api` reports that repository `private: true` (anonymous fetch 404), and `aws-aidp-migrator` is not the name in `plugin.json` (`oracle-ai-data-platform-workbench-aws-migrator`). Plugin commands are namespaced by plugin name, so `commands/inventory.md:14`, `plan.md:14`, `migrate.md:17` and `README.md:51` tell Claude to suggest `/aws-aidp-migrator:plan|migrate|verify` — dead slash commands handed to the user after every verb. `plugin.json` `author.url`/`homepage`/`repository` and `marketplace.json` owner point at the personal repo (siblings point at the oracle-samples tree); `TESTING.md:14`, `docs/DEMO_VIDEO_SCRIPT.md:29` clone it; `docs/build_demo_video.py:162` bakes `/plugin install aws-aidp-migrator` into the shipped video. `ai/claude-code-plugins/README.md`'s plugin table is not updated (the diff touches nothing outside the plugin dir), so a reader of this repo has no working install path at all. Fix: replace every `/aws-aidp-migrator:` with `/oracle-ai-data-platform-workbench-aws-migrator:`, point URLs at oracle-samples, document the repo's `anthropics/claude-plugins-community` flow, add the catalog row, regenerate or drop the video.

### Majors — Athena → Spark SQL transpiler (all executed on PySpark 3.5.9; every case is flags=0 → status ok → verify PASS)

The README says the transpiler "won't corrupt SQL it doesn't recognize" and `verbs.md:36` defines PASS as "auto-translated & runnable". Neither holds today; the wrapper trusts `sqlglot.transpile()` unconditionally and discards its warnings:

- **`array_agg(x ORDER BY y)` → `COLLECT_LIST(x)`** — ordering dropped; **demo query q-1004** (`driver_history_per_policy`) is PASS and returns `['third','first','second']` for events dated Mar/Jan/Feb.
- **`date_add('hour'|'minute'|'second', n, ts)` → `DATE_ADD(ts, n)`** — adds n *days* and returns a DATE (`date_add('hour', 3, …)` → 2024-01-04). **`7/2` → `3.5`** (Presto 3). **`day_of_week`/`dow`/`EXTRACT(DOW)`** → Monday=3 (Presto 1). **`to_hex(sha256(…))` → `HEX(SHA2(…))`** — hex applied twice, 128-char digest (the common PII-hashing idiom; md5 is handled correctly, so the inconsistency is silent). **`log(2, 8)` → `LOG(8, 2)` = 0.333.** **`concat_ws(',', 'x', NULL, 'z')`** → wrapped in a NULL-propagating CASE → NULL (a sqlglot ≥30 regression passed through).
- **`sqlglot` `UnsupportedError` warnings are never read** (`athena_to_spark_sql.py:97` uses the default WARN level): `strpos(s, '-', 2)` loses the occurrence, `max_by(a, b, 2)` loses n, `truncate(1.239, 2)` → `CAST(… AS BIGINT)` = 1, `TRY(CAST('nope' AS DATE))` → bare CAST (raises under ANSI). Re-running with `ErrorLevel.RAISE` catches all four.
- **Multi-statement input is truncated to the first statement** (`[0]` on the transpile result) and reported as a normal rewrite — Athena saved queries routinely hold several.
- **Unknown Presto functions pass through verbatim as PASS**: **demo q-1015** `zip(...)` → `ZIP(...)` → `UNRESOLVED_ROUTINE` on Spark; likewise `format_datetime`, `with_timezone`, `current_time`, `map_agg`, `json_size`, `to_iso8601`, `date_format(ts, '%Y-%m-%dT…')` (`Unknown pattern letter: T`), `CAST(s AS JSON)`, `WITH RECURSIVE`, `> ANY (subquery)`, `approx_percentile(x, w, 0.5)`, `width_bucket(x, ARRAY[…])`.
- **Clauses dropped with no warning**: `LEFT JOIN UNNEST … ON TRUE` → non-OUTER `LATERAL VIEW` (empty-array rows vanish), `split(s, ',', 2)` limit, `FETCH FIRST n ROWS WITH TIES`, `from_unixtime(e, zone)`, `repeat(1, 3)` → `'111'`.
- **`sqlglot>=25.0` is an unsafe floor**: under 25.0.0 the 8 tests still pass but 26 of 113 battery outputs differ — `element_at(arr, 1)` → `TRY_ELEMENT_AT(arr, 0)` (`INVALID_INDEX_OF_ZERO`, would break demo q-1012/q-1016), multi-array `UNNEST` → cartesian product, `sequence()` → `GENERATE_SERIES`. Neither report.json nor the `.spark.sql` header records the sqlglot version. Pin the tested range and stamp it.
- **DDL is not covered**: `CREATE EXTERNAL TABLE … LOCATION 's3://…'` passes through unchanged with the s3 path retained (status ok); CTAS `WITH (format=…, external_location='s3://…', bucketed_by=…)` → `TBLPROPERTIES (… ARRAY('id') …)` which is a Spark `ParseException`; `ALTER TABLE … ADD PARTITION`, `SHOW PARTITIONS`, `PREPARE` parse as opaque `Command` and are still ok.

Fix direction: run the Presto AST after transpile and flag `Order` inside aggregates, sub-day `DateAdd`, `Div` on non-decimal operands, `Hex(SHA2/SHA)`, `ConcatWs`, `Log` arity, `Join`+`UNNEST` non-inner, `Split` limits, `Fetch WITH TIES`, `Command` nodes and `s3://` literals; set `unsupported_level=RAISE`; join all statements; add a Spark-3.5 built-in allowlist check so PASS can mean "runnable"; pin sqlglot. The 8 shipped Athena tests are substring checks that cannot see any of this — add golden-output tests for each case above (and ideally a local PySpark execution test).

### Majors — Glue → PySpark translator (executed; each is flags=0 → PASS unless noted)

- **Every parameterised Glue job is translated to a parameter API the live-validated AIDP toolkit does not use — and it has never been executed on a cluster.** `glue_to_spark.py:169` rewrites `getResolvedOptions` to `args = {"JOB_NAME": oidlUtils.widgets.get("JOB_NAME"), …}` at module scope with no import, classified as a *rewrite* (not a flag) — so all 8 fixture artifacts including both PASS ones carry it (an AST unbound-name scan of the PASS artifacts prints `undefined names: ['oidlUtils']`). The Databricks→AIDP toolkit's live-verified mapping ("dbutils API mapping [verified live: 2026-09-08]") is `dbutils.widgets.* → aidputils.widgets.*` for Python, with `oidlUtils` documented as exposing `.notebook`/`.parameters` (`oidlUtils.parameters.getParameter(name, default)`); `oidlUtils.widgets.get` appears there only as a permissive static-matcher pattern, never as an API use. `scripts/live_land_aidp.sh` has zero mentions of glue, so no translated Glue job has run on AIDP (my own live kernel check was blocked by an expired session token — two independent verifiers rate this major rather than blocker for exactly that reason). If the attribute is absent, every parameterised job raises `AttributeError` on its first executable line while verify says PASS. Fix: emit `aidputils.widgets.get(…)` (or `oidlUtils.parameters.getParameter(k, default)`), fix `README.md:116`, downgrade `resolved_options` to a flag until one translated job has run on a cluster, and add that run to the live script.
- **`write_dynamic_frame.from_options` forces `.mode("overwrite")` and drops `partitionKeys` and `format_options`** (`glue_to_spark.py:249`): a daily incremental sink (Glue appends by default) becomes a full-path wipe with the partition layout lost — destructive, unflagged. **`format="glueparquet"`** is copied verbatim into `spark.write.format("glueparquet")` (not a Spark source; Glue Studio's default sink format); on reads `format_options` (`withHeader`, `separator`) are dropped and only the first of several `paths` is kept.
- **DynamicFrame method-form API is never flagged** (`:307` matches only `<Name>.apply(`): `dyf.apply_mapping(…).resolveChoice(…)`, `drop_fields`, `rename_field`, `glueContext.getSink(…)`, `write_dynamic_frame.from_catalog`, `Spigot`, `DynamicFrameCollection` all survive into a PASS artifact that raises `AttributeError` on a DataFrame.
- **The job-lifecycle regex is receiver-agnostic** (`:132`): `repo.commit()`, `txn.commit()`, `session.commit()`, `db.init()`, `logger.init(cfg)` are all commented out as "removed Glue Job lifecycle" (a JDBC/SQLAlchemy commit silently deleted from a job labelled Ready), while the real `self.job.init(a, b)` is not matched.
- **Multi-line `from_catalog` with non-literal kwargs** (the exact shape Glue Studio generates) embeds newlines from `args.strip()[:60]` in a trailing comment → `IndentationError`; flagged REVIEW, but the artifact cannot be opened by any Python tooling and `push_down_predicate` is silently discarded.
- **verify PASS is a status relabel** (`checker.py:22-28` maps ok→PASS; nothing is compiled or run): a parenthesised multi-line `from awsglue.dynamicframe import (…)` leaves the continuation line → `IndentationError`; `def finish(job): job.commit()` → empty body → `IndentationError`; a Scala Glue job is "translated" as Python with 4 rewrites — all PASS. `README.md:126` ("auto-translates to a runnable Spark script (PASS)") and `html_report.py:73` ("Ready (PASS)") overstate. At minimum `compile()` each artifact and run an unbound-name scan (`glueContext`, `DynamicFrame`, `awsglue`, `Job`, `oidlUtils`, `<transform>.apply`) before granting PASS.

### Majors — plan / verify / report / packaging (executed)

- **Every DCAT table target has a malformed location**: `planner.py:35` does `.replace("/", f"@{ns}/", 1)` on `oci://…`, so the first slash replaced is the one inside `oci://` → `oci:@acme-demo-ns//acme-raw-data/claim_events/` for all 50 tables (0 of 50 well-formed; the Glue translator itself produces the correct `oci://bucket@ns/key`).
- **The `<your-oci-namespace>` placeholder ships inside PASS deliverables**: with `OCI_NAMESPACE` unset (the default `aws-aidp plan inv.json` path; `run_demo.sh` never sets it) plan.json has 56 occurrences, every transfer script 3, and `glue/acme_etl_daily_aggregates.py:23` writes to `oci://acme-curated-data@<your-oci-namespace>/…` while verify says PASS. `cmd_plan` never calls `load_dotenv()` (only `cmd_inventory` does), so `README.md:66`'s "put OCI_NAMESPACE in .env" does nothing for plan/migrate. Fail fast or flag.
- **`migrate` crashes with `UnicodeEncodeError` on Windows default encoding** (`runner.py:193` `write_text` without `encoding=`; the `→` comes from the tool's own finding text) — exit 1 after report.json, no report.md/html; `html_report.render` is additionally wrapped in `except Exception: pass`. Every Windows plugin user without `PYTHONUTF8` hits this. Pass `encoding="utf-8"` everywhere and log render failures.
- **Duplicate names overwrite artifacts while both rows report ok** (`runner.py:32`, paths keyed on the human name; Athena permits identical NamedQuery names across workgroups — the fixture models two): two `dup_name` queries → one `.spark.sql` containing only the second, verify PASS for both. Asset names are also used unsanitised as paths — `../../traversal_probe` wrote outside `-o` (`runner.py:60`).
- **`verify --filter glue` hides the 52 catalog stubs** (migrate filters on `source_type`, verify on target `kind`): demo.sh's headline verify prints `SKIP: 0` while report.json holds 52 `planned` Glue assets. And verify never inspects artifacts: deleting `athena/ glue/ transfer/` or corrupting every `translated_sql` still yields PASS 20.
- **Generated rclone config has an inline `# comment` on the `env_auth = true` line** (`s3_to_oci.py:72`); rclone's INI parser (goconfig) only treats `#`/`;` at line start as comments, so `env_auth` becomes `true          # uses AWS_PROFILE …` and the S3 remote fails to construct; the `mkdir || true` masks its own failure and the copy then fails. Code-read against fetched docs + parser source (rclone not installed here). Compartment OCID and OCI region are never wired either (`build_transfer` accepts them, the runner never passes them; every script hard-codes `<compartment-ocid>` and `us-ashburn-1`).
- **License metadata contradicts itself**: `pyproject.toml:10` `license = { text = "Apache-2.0" }` while LICENSE, both plugin.json files and both PR bodies say MIT; the built wheel's METADATA reads `License: Apache-2.0` + `License-File: LICENSE` (MIT text). `scripts/build_pr_trees.sh:56-61` explains it — the packaging script patches plugin.json to MIT but never pyproject. Siblings also ship `NOTICE` and `PRIVACY.md`; this plugin ships neither, and it needs a PRIVACY.md because `scripts/ai_assist_flags.py` sends flagged source SQL/PySpark to an LLM (see below).

### Majors — live path and the live-validation claim (executed with fakes/Stubber; nothing touched AWS or OCI)

- **"Verified end-to-end on a live AIDP workspace" is not exercised by any migrator code path.** The only live artifact, `scripts/live_land_aidp.sh`, uploads a 5-row CSV written inline in the script with `oci os object put` (not the generated rclone job) and runs a **hand-typed** Spark query (`:49-52`) whose text differs from what the translator emits for the seed query (different table, date predicate dropped, `split(…, ';')` vs the transpiler's `SPLIT(…, CONCAT('\\Q', ';', '\\E'))`). Non-demo `migrate` returns `skipped: live mode not yet wired` for all Athena/Glue assets with zero AIDP calls, and `AidpClient`/`run_until_green` are referenced by nothing outside their own modules. The PR body, pptx slides 1/8/11 and `SPEAKER_NOTES.md:61-63` say "proven end-to-end"/"Working"; `README.md:153`, `SKILL.md:43/56`, `TESTING.md:141-142` and `SESSION_RUNBOOK.md:176` say "demo-only". Pick one truth. Likewise **"Glue Data Catalog — working"** (`SKILL.md:54`, README mapping table, slide 11): all 2 databases + 50 tables are `SKIP translator not yet implemented` stubs; no DDL is produced.
- **`scripts/live_teardown.py` deletes by fixed name with no tag/ownership check, confirmation or dry-run** (`:42`): recorded calls against a fake session — `delete_database('acme_curated')` (drops *every* table in it), `delete_named_query` for **every** query the workgroup lists (including two non-seed ids I injected), `delete_work_group(RecursiveDeleteOption=True)`, empty + `delete_bucket`. The `Project=aws-aidp-migrator-test` tag the seed applies is never read back; `AWS_PROFILE=<prod> python3 scripts/live_teardown.py` deletes immediately.
- **Inventory swallows per-item `AccessDenied`** (`glue.py:76-83`, `athena.py:33/55`, same pattern in emr/sagemaker): with `get_tables(db2)` and `get_jobs` denied, the summary is `table_count: 1, job_count: 0` with no error key, so a partially-authorised scan is presented as the complete estate and plan/migrate proceed on it.

### Minors (executed unless noted)

`aws-aidp --version` prints 0.1.0 (pyproject/plugin.json/CHANGELOG say 0.2.0) and stubs say "deferred to v0.2" in the 0.2.0 release; CHANGELOG says 15 tests (19), describes a regex UNNEST fix that has no counterpart in the sqlglot code, and cites `docs/FDE_PRESENTATION.md` which doesn't ship; README lists `DEMO.md` (absent), `TESTING.md:20` says boto3 is the only runtime dependency (sqlglot is a hard import); `--filter sagemaker` matches nothing (source types are `sm_*`) and `--filter` accepts any string silently; `--sources` is ignored in fixture mode; `report.html` is written but never mentioned by the CLI/verbs.md/TESTING; the `_NAMED_REWRITES` substring check mis-attributes rewrites in the report (`json_extract` listed for a query that only uses `json_extract_scalar`); `.toDF()` is stripped from RDDs (`rows.toDF()` → `rows`), a positional `write_dynamic_frame.from_options(dyf, …)` invents a `df` receiver, a non-literal `getResolvedOptions` list becomes `args = {...}` (a set of Ellipsis), JDBC/DynamoDB `from_options` are replaced with a parquet stub (connection details deleted from the artifact), `s3://` inside comments/docstrings is rewritten, `_kwarg` truncates at apostrophes and `_balanced_args` mis-handles `\\"`, `\w+\.spark_session` rewrites unrelated objects; `RunStore` writes under `~/Documents` by default with unsanitised keys; `AidpClient` supports api_key auth only (session-token profiles → `KeyError 'user'`; the toolkit supports both); `list_buckets` ignores `ContinuationToken`; no test touches any live/cloud module; `demo.sh`/`run_demo.sh` embed `/tmp` inside inline Python and assume `python3` (fails under Git Bash); `report.md` code fences can be broken out of by SQL containing ```` ``` ````; the fixture builder stamps wall-clock `scanned_at`, so regenerating dirties the tree; `requires-python >=3.9` cannot be satisfied with the `mcp` extra (≥3.10). Live-path minors (verifiers downgraded these from major; the code is unreachable from the CLI today): `demo.env.example` defines `AIDP_*` names with an `export ` prefix while `AidpConfig.from_env` requires `OCI_PROFILE/OCI_REGION/DATALAKE_OCID/WORKSPACE_ID` and `README.md:66` points at a non-existent `.env.example`; `run/jobruns.py:115` fires a repair and immediately re-polls, so a 202-accepted repair followed by a stale FAILED read burns all three repairs in 10 ms and reports FAILED; `live_seed.py:163` turns `AlreadyExistsException` into `update_table`/`update_job`, repointing a same-named customer table/job at the demo CSV; `scripts/ai_assist_flags.py` base64-embeds the full flagged Athena SQL / Glue script into `ai_generate('google.gemini-2.5-flash', …)` on the user's own cluster (not Anthropic — the `anthropic` extra is dead code) and rewrites `report.json`/`report.html` in place, persisting even `(ai_generate call failed: …)` as a suggestion; nothing discloses that egress (hence PRIVACY.md), `--help` exits 1 unless `AIDP_DATALAKE` is set, and it hard-depends on the sibling engineer-agent plugin at a version-pinned cache path (`demo.env.example:24`) that `plugin.json` never declares as a dependency; `_balanced_args` treats an escaped backslash before a closing quote as escaping the quote, so one `"x\\"` argument swallows the rest of the file into the call (verifiers rate this above note).

### Notes — repo hygiene and content

- **Internal material shipped in a public sample**: `docs/LIVE_DEMO.md:126/130` and the `scripts/ai_assist_flags.py:13` docstring name an Oracle-internal AIDP workspace nickname and its Spark cluster (values deliberately not reproduced here); `DEMO_VIDEO_SCRIPT.md:119` and `TESTING.md:151` name colleagues and an internal channel; `SESSION_RUNBOOK.md`/`SPEAKER_NOTES.md`/`LIVE_DEMO.md` are first-person presenter scripts ("Ask me to start it", `cd ~/Documents/aws-aidp-migrator`) — while `SESSION_RUNBOOK.md:5` asserts "No personal names anywhere — safe to share". `scripts/build_pr_trees.sh` is the tooling that generated these PR trees (rsyncs from a personal checkout into `~/Documents/aidp-samples-prs`) and `docs/build_*.py` are deck/video generators with undeclared matplotlib/PIL/python-pptx deps; the 3.7 MB of mp4/gif/pptx/pdf/png binaries have no precedent under `ai/` (largest tracked file today is 509 KB). Suggest keeping the presenter kit and generators in your internal repo and shipping only user docs.
- The demo counts in `LIVE_DEMO.md:91/113`, `SESSION_RUNBOOK.md:126` and the video's verify card (18/2, 20/8/85) are three different sets, none matching the tool's output (20/13/80); the video's footer renders the `→` as a missing-glyph box on every card (font at `build_demo_video.py:42-44`), and `DEMO_VIDEO_SCRIPT.md` describes a 5–7 min screen recording while the mp4 is a 2:05 TTS card slideshow.
- `README.md:5-7` and slide 3 state an unsourced market statistic ("about half of incoming AIDP migrations…") — not appropriate in an Oracle-published sample; "AIDP DCAT"/"AIDP saved queries" are not AIDP vocabulary (0 hits in the toolkit and the 37 engineer-agent skills; the AIDP term is catalog/schema/table and "verified queries"), and plan targets omit the catalog level entirely.
- Merge gate you already know about: OCA unsigned (CI red), no `Signed-off-by` trailer, PR template/issue reference not used — CONTRIBUTING.md requires all three.

**Bottom line:** keep the architecture (four verbs, sqlglot AST core, generated rclone job, flag-don't-guess) — it is the right shape, and the offline demo proves it. Before this can merge: pin `mcp<2`, make the plugin self-installing via `${CLAUDE_PLUGIN_ROOT}`, fix the plugin identity/namespace and register it in the catalog, settle the widgets API with one live run of a translated Glue job, stop `sqlglot` warnings from becoming PASS, fix the planner location bug and the Windows encoding crash, align the license metadata, add PRIVACY/NOTICE, make the live claims match the code, and move the presenter kit out of the public tree. Happy to re-verify. 🤖 Deep review by Claude (Fable)

mppotnuru:
Thanks for the exceptionally thorough review — force-pushed a substantial revision that addresses it.

**Engine (both PRs):** the Athena transpiler was reworked off `sqlglot` onto a hardened deterministic translator with explicit semantic-gap guards, so the silent-semantics findings (`array_agg ORDER BY`, sub-day `date_add`, integer `/`, `day_of_week`, `to_hex(sha256)`, `log` arity, `concat_ws` NULLs, unknown-function passthrough, multi-statement truncation, sqlglot version drift) are gone — those cases now flag rather than PASS, backed by a corpus + Spark-runtime tests. Glue now emits `oidlUtils.parameters.getParameter(...)`, flags method-form DynamicFrame APIs, and `verify` PASS requires a compile + unbound-name scan.

**Fixed here:**
- MCP: `mcp>=1.2,<2` (+ py>=3.10 marker); `stdin=DEVNULL` (Windows hang); self-installing `.mcp.json` (`python -m` + `${CLAUDE_PLUGIN_ROOT}` / cwd for Codex); real ImportError surfaced.
- Packaging: `fixtures/` now under `aws_aidp.fixtures` (packaged); planner OCI location bug fixed; Windows `UnicodeEncodeError` fixed (utf-8); collision-safe artifact paths; namespace validated.
- Identity: plugin name/namespace/URLs point at oracle-samples; private-repo + dead slash-command refs removed; broken links fixed.
- Licensing/hygiene: MIT aligned across pyproject/LICENSE/plugin.json; added `NOTICE` + `PRIVACY.md`; rclone inline-comment fixed; unsourced market stat + non-AIDP "DCAT" vocabulary removed; presenter kit, live scripts, build tooling and demo media removed from the shipped tree; no internal names/OCIDs.
- Process: `Signed-off-by` on all commits (OCA green); registered in the plugin catalog / marketplace manifests.

Remaining follow-ups called out honestly: one live run of a translated Glue job on a cluster to promote the widgets mapping from flagged→rewrite, and the shared-plugin-content dedup between the two trees. Happy to iterate. 🙏

mppotnuru:
Update (additive commit, no force-push): the **"verify PASS is a status relabel"** major is now addressed.

Root cause was exactly as you found — when the SQL side moved off `sqlglot` to regex rules, it lost the free parse-validation the transpiler gave for granted (`parse_failed` finding count went 1 → 0). Restored it as explicit **output-validation gates** in `athena_to_spark_sql.translate()`, mirroring the `ast`-parse gate the Glue translator already had:

- **multi-statement**, **unrecognised/unbalanced statement**, and **10 Presto-only constructs** Spark rejects (`WITH RECURSIVE`, `FETCH FIRST`, `CAST AS JSON`, CTAS `WITH(...)`, `> ANY(...)`, `AT TIME ZONE`, `TABLESAMPLE`, `PREPARE/EXECUTE`, `UNLOAD`) → now flag → REVIEW instead of PASS.
- A **Spark-3.5 builtins allowlist** so unknown functions are caught rather than passed through.
- Verdict wording made honest across report.html / verbs.md / README: **PASS = "translated, no known issue detected — not execution-verified."**
- Adversarial 20-query junk manifest that previously returned PASS 20/20 (incl. `CREATE EXTERNAL TABLE … s3://…` byte-identical) now flags correctly; added `test_output_validation.py` + `test_verdict_claims.py`.

Also included the earlier Spark-3.5.3 translator fixes (JOIN-after-UNNEST, 1-based subscripts, scoped `.spark_session`). Full stress suite green. Thanks again — happy to re-verify.

ahmedawan-oracle:
## Re-review of `d962dff` — the fix round lands most of what was asked (MCP, packaging, identity, hygiene, verify contract, a 370-test suite); the regex rewrite closes the round-1 silent-drift list but opens four new PASS-then-wrong classes, two claimed fixes did not land, the plugin's skill currently loads with empty metadata, and the PR body/CHANGELOG/TESTING still describe the old plugin

@mppotnuru — thank you for the scale of this revision; a lot of it is verified working below. Method as before, by execution on the new head (and on #109's `aa46d16`, whose engine tree is byte-identical): the full suite, a 155-case Presto battery **executed on PySpark 3.5.9**, the round-1 hostile Glue cases plus 43 new ones, the MCP server over stdio with a real client, a non-editable-install simulation, `claude plugin validate` **and** an actual `claude --plugin-dir` load, the published `20260430` swagger, and every doc claim against the code. Every finding below survived independent adversarial re-verification.

**Suite as shipped:** `pytest tests` → 370 collected, **360 passed, 2 failed, 8 skipped** on both trees. The two failures are yours to fix, not environmental: `tests/test_translator_followups.py::test_teardown_tracks_failures_and_exits_non_zero` reads `scripts/live_teardown.py`, which this same revision deleted (the test can never have passed); `tests/stress/test_scale.py::test_one_megabyte_clean_inputs_do_not_hang` asserts < 10 s and measured 11–27 s in every run here (see the quadratic-cost item). `TESTING.md` tells testers `pytest -q` is the pass criterion.

### Verified fixed ✔

- **MCP**: `mcp>=1.2,<2` resolves to 1.30; `stdin=DEVNULL` — all four tools return in < 1 s on Windows over stdio, full pipeline via MCP = PASS 13 / REVIEW 20 / SKIP 80; #108's `.mcp.json` shape (`python -m aws_aidp.mcp_server`, `PYTHONPATH=${CLAUDE_PLUGIN_ROOT}`) works from an unrelated cwd with the system interpreter; fixture mode works without boto3 and live mode fails with a clean `boto3 is required`.
- **Packaging**: fixtures ship inside the package (package-dir-only install runs the pipeline from a neutral cwd; `--version` 0.2.0); planner locations 50/50 well-formed `oci://<bucket>@<ns>/<key>`; no `UnicodeEncodeError` under cp1252 and `report.html` is written; duplicate names → distinct sha-suffixed artifacts; `../../x`, `..\..\y`, `CON` stay inside `out_dir`; `verify` fails closed on all 7 tamper cases (missing artifact, count mismatch, missing `complete`, in-progress marker, ok-with-flags, absolute/escaping `output_path`, bogus filter); non-demo `migrate` now refuses with exit 2; placeholder namespace can no longer reach PASS (flagged `oci_namespace_missing`); HTML/markdown escaping, fence growth and header injection all neutralised; license MIT everywhere with `license-files`; NOTICE + PRIVACY.md present; `Signed-off-by` on all commits; catalog row added.
- **Athena translator — every round-1 silent case now flags**: `array_agg ORDER BY`/`DISTINCT` (demo q-1004 is REVIEW), sub-day `date_add`, `day_of_week()`/`dow()`, `to_hex(sha256(…))`, `strpos/3`, `max_by/3`, `truncate`, `TRY()`, multi-statement, `FETCH … WITH TIES`, `from_unixtime(tz)`, `LEFT JOIN`/`FROM`/multi-array/map `UNNEST`, `WITH ORDINALITY`, `UNNEST` followed by `JOIN`, `arr[1]`/`arr[i]`, DDL with `s3://` `LOCATION`, CTAS `WITH(…)`, `UNLOAD`/`PREPARE`/`WITH RECURSIVE`/`> ANY`/`AT TIME ZONE`, and unknown functions (`format_datetime`, `with_timezone`, `map_agg`, `json_size`, `listagg`, UDFs) — the allowlist equals the live `SHOW FUNCTIONS`. `zip` → `arrays_zip` is a correct rewrite (q-1015 executes). `concat_ws` and `log(2,8)` PASS legitimately (Spark matched Presto on execution). Rewrites never touch literals or comments; output is idempotent; all 31 README coverage rows reproduce; `test_spark_runtime` executes real semantics (14 pass under `AIDP_SPARK_VERSION=3.5`); PASS is worded honestly in CLI, report.md, report.html, README, verbs.md.
- **Glue translator — every round-1 defect fixed**: `getResolvedOptions` → the documented `oidlUtils.parameters.getParameter(name, default)`; no forced `.mode("overwrite")`, `partitionKeys` → `.partitionBy(…)`, `glueparquet`/`format_options` flagged; method-form DynamicFrame APIs, `getSink`, write-from-catalog, `Spigot` flagged; lifecycle regex bound to `Job()`-assigned receivers; multi-line `from_catalog` compiles; `.toDF()` preserved; positional write frame flagged; non-literal `getResolvedOptions` flagged; JDBC/DynamoDB left in place; comments/docstrings untouched; argument parsing is `ast`/`tokenize`-based; `.spark_session` scoped; Scala jobs flagged; parenthesised multi-line imports roll back via the `ast.parse` guard (the translator can no longer emit broken Python); alias imports detected; no over-flagging of plain PySpark; idempotent on all 71 cases.
- **Live path**: client matches the published `20260430` swagger (path, `jobKey` body, `Parameter[]`, 201, repair body, key limits); retries with a reused `opc-retry-token`; the full `State.status` vocabulary (unknown status fails closed instead of hot-polling); `taskRunSummaryMap` parsing; inventory `AccessDenied` now produces per-item warnings that `summarize()` prints; `.env.example` names match `require()`; live scripts, presenter kit, build tooling and all binaries removed; internal names gone; fixture JSON-identical to round 1.

### Majors — Athena regex translator

Moving off `sqlglot` removed the parser that gave you literal handling, type syntax and linear cost for free. The flag rules now catch the round-1 list, but a regex pass has no notion of the string it is rewriting, and that shows up as a new silent class (all executed on Spark 3.5.9, all `flags=0` → `ok` → PASS):

1. **[NEW] `date_format` literal quoting emits Trino `''` escapes Spark cannot read.** `_translate_strftime` wraps bare letters in quotes and `_rule_date_format` (`:281`) re-escapes them Trino-style, so `date_format(ts, '%Y-%m-%dT%H:%i:%sZ')` → `'yyyy-MM-dd''T''HH:mm:ss''Z'''` with 0 flags; Spark concatenates adjacent literals and fails with `Unknown pattern letter: T` (`'Day %d of %M'` → `Unknown pattern letter: o`). `tests/stress/test_athena_corpus.py:108-114` asserts the broken string as expected output, and `test_spark_runtime` only executes `'%Y-%m-%d'`. Emit `\'` escaping (or double-quote the Spark literal) and add a runtime test with a literal letter.
2. **[NEW] Trino string-literal escaping passes through unchanged — silent wrong values.** No rule touches literal content: `'it''s'` → Spark returns `its`; `length('a\nb')` → 3 (Presto 4); `'C:\temp\new'` acquires a TAB and a NEWLINE. `WHERE surname = 'O''Brien'` silently matches `OBrien`. The sqlglot head emitted `'it\'s'`/`'\\n'`; this is a regression. `_masked_sql` already recognises `''`, so the translator knows where these literals are — add a first rule that rewrites `''` → `\'` and doubles backslashes (or flags literals containing either when the target parser config is unknown).
3. **[NEW] `ARRAY[…]`, `MAP(ARRAY…)`, `array(varchar)`/`map(…)`/`row(…)` types PASS with 0 flags but Spark cannot parse them.** `ARRAY[1,2,3]`, `ARRAY['a','b'][1]`, `approx_percentile(x, ARRAY[0.5,1.0])`, `width_bucket(7, ARRAY[1,5,10])`, `MAP(ARRAY['a'], ARRAY[1])`, `CAST(NULL AS ARRAY(VARCHAR))`, `CREATE TABLE t (b array(varchar), c map(varchar,bigint))` → `ParseException`/`UNRESOLVED_COLUMN`; the sqlglot head produced `ARRAY(...)`, `MAP_FROM_ARRAYS`, `STRUCT<>`. `_rule_array_subscript:1141` even comments "the unsupported construct rule already reports it" — nothing does. Add a gate for `\bARRAY\s*\[` and the `array(`/`map(`/`row(` type forms (or a safe `ARRAY[` → `array(` rewrite with bracket matching).
4. **[NOT FIXED] Integer division still PASSes silently** although your fix list names "integer `/`": `7/2` → 3.5, `count(x)/count(*)` → 0.667 (Presto 0), `SUM(a)/COUNT(*)` → 1.5 (Presto 1); the only `/` in the module is in `_DANGLING_TAIL`. Ratio-of-counts is routine analytics. Flag every `/` whose operands are not provably non-integer, and add the README row.
5. **[NOT FIXED] `EXTRACT(DOW FROM x)` still silent** — `_UNSUPPORTED_PATTERNS:974` matches `day_of_week(`/`dow(` only; the standard-SQL form returns 2/1 on Spark for Monday/Sunday (Presto ISO 1/7). Extend the pattern to `EXTRACT\s*\(\s*(DOW|DAY_OF_WEEK)\s+FROM`.
6. **[NEW] `greatest()`/`least()` NULL semantics differ silently** — both are Spark built-ins so they pass every gate; `greatest(1, NULL)` → 1 on Spark, NULL in Athena. Add them to a null-semantics flag rule like `array_agg_semantics`.

Minors (executed): **quadratic cost** — `_matching_paren`/`_top_level_argument_spans` recompute `_masked_sql` for every matched call (400 `date_format` calls 8 s, a 26 KB query with 800 calls 38.7 s, the 1 MB `test_scale` input 21–27 s → the shipped test fails; Athena saved queries can be 256 KB; cache the mask per pass); **over-flagging of valid Spark SQL** — `GROUPING SETS` (`unknown_function 'sets'`), `CASE WHEN … THEN (x) ELSE (y)` (`'then'`, `'else'`), `(SELECT …) UNION ALL (SELECT …)` (`statement_not_recognized`), `FROM (VALUES (1)) t(x)` / `UNNEST(arr) u(x)` (`'t'`, `'u'`), `SELECT 'p' AS json` (`CAST AS JSON` regex), `SELECT 1 AS div` (`_DANGLING_TAIL`), a 120-char `complex_interval` span — GROUPING SETS and CASE were zero-flag on the sqlglot head; **still PASS-but-fail on Spark** — paren-less `current_time`/`localtime`, `TIME '12:34:56'`, 2-arg `IF`, weighted `approx_percentile(x, w, 0.5)`; **remaining silent drifts** — `repeat(1,3)` → `'111'` (on your own list), `random(10)`, `LIKE 'a\_b'` without `ESCAPE`; one construct yields two flags (`athena_datetime` + `unknown_function` for all six datetime functions; AS-less `UNNEST` aliases); `_rule_array_subscript`'s docstring says "Rewrite … to element_at" but it only flags; `report.json`/`.spark.sql` still carry no tool or Spark-target version (round-1 ask; `SPARK_VERSION` exists in `spark_builtins.py` but is never written out); `scripts/generate_spark_builtins.py`, referenced by the allowlist header and `test_spark_runtime`, is not in the tree.

### Majors — Glue translator

- **[NEW] GlueContext-only methods survive as `spark.<method>` with 0 flags → PASS, `AttributeError` at runtime.** After `glueContext = GlueContext(sc)` → `glueContext = spark`, `create_data_frame.from_catalog`/`create_data_frame_from_catalog`, `forEachBatch`, `get_logger`, `extract_jdbc_conf`, `purge_table`, `purge_s3_path`, `start_transaction`/`commit_transaction`, `write_from_jdbc_conf`, `create_sample_dynamic_frame_from_catalog`, `add_ingestion_time_columns` are neither rewritten nor flagged (`_RESIDUAL_GLUE_API_PATTERN:2383` matches only `.create_data_frame(`). A Glue streaming or Lake Formation job is graded PASS and dies on its first such call. You already track GlueContext receivers in `_rule_context_bootstrap` — flag any attribute call on them that is not a rewritten entry point.

Minors (executed): **the "verify PASS requires a compile + unbound-name scan" claim does not hold** — nothing in the tree compiles artifacts; a report whose 0-flag `glue/broken.py` contains `def (:` is PASS, exit 0 (the translator's own `ast.parse` guard is what prevents broken output; verify itself has no gate — either drop the claim or add the gate); **`JOB_NAME` is not special-cased** — every `getResolvedOptions` key including the Glue-injected `JOB_NAME` becomes `_aws_aidp_required_parameter("JOB_NAME")` classified as a *rewrite*, so `test_clean_job_has_no_flags` asserts PASS for `['JOB_NAME']` and all 8 fixture artifacts raise `missing required AIDP job parameter: JOB_NAME` at startup under a stub kernel that supplies only `run_date`; your comment says the mapping "stays flagged until a live run" — it is not flagged; a Glue-signature `join(paths1=, paths2=, frame2=)` on a dict-held frame passes with 0 flags (`_DATAFRAME_COMPATIBLE_METHOD_NAMES:1159` needs lineage that subscripts never get); a user-defined `class Job` has its `init()`/`commit()` calls replaced by `pass` before `residual_glue_api` flags the class (user logic deleted, REVIEW but lossy); plain PySpark `sc = SparkContext(appName=…)` is routed into the Glue pipeline by the `\bSparkContext\b` gate and rewritten with `appName` dropped; README Glue rows `.toDF() → collapse to df` and `any s3:// literal → oci://` no longer describe the code (`.toDF()` is now preserved; `s3://` outside a recognised I/O call is *flagged*, not rewritten) — a doc regression from this revision; the bundled demo now yields **zero** PASS Glue artifacts (`job_bookmark` fires on 7 of 8 fixtures, `DropNullFields` on the 8th) — fine as fail-closed policy, but the demo no longer shows the PASS path.

### Majors — plugin packaging and install

- **The plugin's skill and one command load with empty metadata in both plugins.** `skills/aws-aidp-migrator/SKILL.md` frontmatter is not valid YAML — the unquoted `description:` contains `(verbs: inventory, …)`, a `: ` inside a plain scalar (`ScannerError: mapping values are not allowed here`); `commands/inventory.md`'s `argument-hint` starts with `[` and parses as a flow sequence. Executed with Claude Code 2.1.263: `claude plugin validate` on a copy without `marketplace.json` reports "YAML frontmatter failed to parse … loads with empty metadata (all frontmatter fields silently dropped)", and a real `claude --plugin-dir` load logs `[WARN] Failed to parse YAML frontmatter` for both files — so the skill's `description` (its discovery trigger) and the command's `allowed-tools`/`argument-hint` are dropped at runtime. Quote both values. This is pre-existing (same text in the round-1 head); my first review missed it because the validator only checks `marketplace.json` when one is present — apologies for that.
- **[NOT FIXED, partly regressed] The README's Claude install flow cannot work.** `README.md:48` says `/plugin marketplace add oracle-samples/oracle-aidp-samples` then `/plugin install oracle-ai-data-platform-workbench-aws-migrator` — the repository has no root `.claude-plugin/marketplace.json` (executed with the `claude` CLI: the marketplace add fails), and the repo's documented flow is `anthropics/claude-plugins-community`. Meanwhile the plugin-local `.claude-plugin/marketplace.json` still has the author's personal owner and its plugin entry was **renamed in this revision** from the long name (which matched `plugin.json`) to `aws-aidp-migrator`, so even that manifest no longer serves the documented install name. Follow the sibling pattern (ask-aidp: Oracle owner, entry name == `plugin.json` name, `source: "./"`) or drop the file, and document the community-marketplace flow.
- **The PR body still describes the previous plugin**: "Athena→Spark SQL via **sqlglot** (AST transpiler, not regex)" (sqlglot is gone: `pip show` → `Requires: boto3`; the translator is regex rules), "Verified end-to-end on a live AIDP workspace" (non-demo `migrate` exits 2 `live AIDP migration is not implemented`; no verb reaches `AidpClient`), "19 unit tests" (370). GitHub shows the body was never edited after the force-push. The body is the merge record — please rewrite it.

### Minors — pipeline, MCP, live path (executed)

`demo.sh` still crashes under Git Bash at the inline `python3 -c "…open('/tmp/…')"` step (rc 1 after all four verbs succeed; `TESTING.md` §2 makes exit 0 the pass criterion); `cmd_plan` still never calls `load_dotenv()`, so `.env`'s `OCI_NAMESPACE` is ignored and the plan carries the placeholder (README's "Real AWS" recipe implies `.env` is honoured); `_env.py:13` reads `.env` without `encoding=` — a non-ASCII byte crashes `inventory` under cp1252 (`'charmap' codec can't decode`) and UTF-8 values become mojibake, so "UTF-8 everywhere" stops at the outputs; MCP tool failures are still plain text with `isError=False` (`[exit 2]`), including the new timeout path; `.mcp.json` runs `python` (absent on stock macOS 12.3+ and Debian without `python-is-python3`) and needs the `[mcp]` extra that the plugin section of the README never mentions (`pip install -e .` only); the `ImportError` cause is still masked ("real ImportError surfaced" did not land — `SystemExit(str)` never prints the chained cause); `migrate`'s summary line still omits `skipped=` and `report.html`; S3 transfer scripts are the one artifact written non-atomically and outside stale-artifact detection; `list_buckets` still ignores `ContinuationToken`; `RunStore` still defaults to `~/Documents/aws-aidp-migrator/state` with no override. Live path (unreachable from the CLI today, so minor): the repair loop still re-enters `poll_until_terminal` immediately after a 201 repair and burns the whole budget on stale FAILED reads; a CANCELED run with a failed task is still repaired; the signer is still API-key only (`KeyError 'user'` on an `oci session authenticate` profile — the toolkit branches to `SecurityTokenSigner`); `get_job_run` never sends `shouldIncludeTaskRunSummaries`, so a FAILED run without summaries yields `failed_tasks=[]` and no repair; `AIDP_ENDPOINT` collides with the sibling plugins' host-only convention (`export AIDP_ENDPOINT="https://aidp.<region>.oci.oraclecloud.com"` makes `from_env()` raise); the default host `datalake.<region>` matches the swagger template and resolves to the same addresses as the toolkit's live-tested `aidp.<region>` gateway, but no authenticated call has been made on the new route — one `GET <ws_base>/jobs` on a real tenancy, recorded in the PR, would settle it.

### Docs that still describe the old plugin

`CHANGELOG.md` is byte-identical to the round-1 head: it lists the deleted `live_seed.py`/`live_teardown.py`, `docs/FDE_PRESENTATION.md`, `docs/MCP.md`, "11 deterministic rules" (13 now), "15 standalone tests" (370), and has no entry for the sqlglot→regex rewrite, the gates, the stress suite or the 09-12/09-14 commits. `TESTING.md` §6 still tells testers to run the deleted live scripts, still has the stray second `cd aws-aidp-migrator` (also in README §Quick start — a copy-paste of the quick start fails before `pip install`), still "(for interns)", and claims a "CI compatibility matrix 3.9–3.14 plus a native Windows run" — the shipped `.github/workflows/stress-tests.yml` sits inside the plugin directory, where GitHub Actions never reads it (repo root `.github` has only the PR template; neither PR touches it), so the README's "run in CI" is also untrue. README: links `docs/MCP.md` (deleted in this revision — a link that was valid before), lists `DEMO.md`, still says `/aws-aidp-migrator:inventory` (commands were fixed to the long namespace, the README was not), still "OpenAI's CLI has no plugin marketplace" (this repo ships one and #109 registers in it), "Glue Data Catalog … → AIDP catalog (schemas/tables)" and SKILL.md "Glue Data Catalog — working" while all 52 catalog assets are `SKIP translator not yet implemented; deferred to v0.2` at version 0.2.0, `verbs.md`/`commands/plan.md` list 6 of the 9 target types and `--filter athena|glue|sagemaker` of the 5 values, the "Compatibility contract" documents live-call defaults for a path no verb can reach, `pip install -e '.[oci]' # for live AIDP calls` likewise, the awsglue-import row says "commented out" while the code emits `pass` and drops the import text, SKILL.md still offers `pip install aws-aidp-migrator` (PyPI 404), PRIVACY.md describes an "AI-assisted rewrite step" that was removed, NOTICE omits the `oci`/`requests` extras, pyproject keeps the dead `llm = ["anthropic>=0.40"]` extra, and the "other half of customers" framing remains.

**Bottom line:** the packaging, MCP, verify-contract and hygiene work is solid and verified, and the flag-first policy did what it set out to do for the round-1 list. What stands between this and merge is now narrow and specific: the three literal/escaping/`ARRAY[]` regressions and the two unfixed drifts in the Athena engine (or bring `sqlglot` back as the parser underneath the flag rules — it handled exactly these), the GlueContext-method gap, the unparseable frontmatter (the skill is invisible today), a working install path plus a coherent `marketplace.json`, the two red tests, and a PR body/CHANGELOG/TESTING/README pass that describes the plugin you actually shipped. Happy to re-verify. 🤖 Deep review by Claude (Fable)

navid-heydari (commented):
Thank you Mounika for the substantial follow-up work. The latest changes resolve many earlier concerns, including MCP startup, packaging, and fail-closed validation.

Before merge, please address the remaining cases that can still produce invalid or semantically different Spark output, complete the GlueContext detection, fix the plugin metadata and install-path mismatch, and return the committed test suite to green. Please also update the PR description and documentation to match the current implementation and supported workflow.

Just as a side note/Food for thought: Consider the scale of the migration as well, in another word, in case of having large number of tables and views and complex flow between the domains, how much does it cost for customer to use the plugin for migration. AI token wise and data transfer wise.

### P1 — Fix invalid `date_format` output

`date_format(ts, '%Y-%m-%dT%H:%i:%sZ')` is translated with Trino-style quote escaping, which Spark cannot parse. Emit Spark-compatible escaping or flag these formats, and add a runtime regression test.

### P1 — Flag remaining unsafe Athena constructs

`ARRAY[...]`, integer division, `EXTRACT(DOW ...)`, and `greatest`/`least` with NULL values currently receive zero flags despite Spark incompatibility or semantic drift. Add validation gates or safe rewrites.

### P1 — Flag unhandled GlueContext methods

After `GlueContext` is rewritten to `SparkSession`, methods such as `get_logger`, `start_transaction`, and `create_data_frame_from_catalog` remain executable but unsupported. Flag unhandled calls on tracked GlueContext receivers.

### P1 — Repair plugin frontmatter

The skill description and inventory command’s `argument-hint` are invalid YAML, causing Claude to drop their metadata. Quote these values and validate all skill and command frontmatter in tests.

### P1 — Restore a green test suite

A committed test reads the deleted `scripts/live_teardown.py` and fails with `FileNotFoundError`. Restore the supported workflow or remove the obsolete test and update the test documentation.

### P2 — Align plugin installation metadata

The marketplace entry and documented install name do not match `plugin.json`. Align the plugin name, Oracle ownership, and README installation flow, then validate it from a clean environment.

### P2 — Update PR and user documentation

Refresh the PR body, README, CHANGELOG, and TESTING guide to reflect the current translator, supported migration workflow, and test suite.

lucas-de-almeida-reis (commented):
Coming at this from the catalog side rather than the code — @ahmedawan-oracle's two rounds cover
correctness exhaustively and I'm not adding to that list. One observation about framing that I
think sits underneath several of the findings already raised.

## The doc inaccuracies have a single cause

The review asks for "a README pass that describes the plugin you actually shipped", and lists the
places where it doesn't: `.[oci]` advertised "for live AIDP calls" on a path no verb reaches, the
Glue Data Catalog documented as working while all 52 catalog assets return
`SKIP translator not yet implemented`, the PR body claiming end-to-end verification on a live
workspace.

Those aren't independent slips. They follow from the artifact presenting itself as a migrator
while what it implements is, right now, closer to a translator or assistant for migration. Three more that haven't been raised, all the same shape:

- **`README.md:3`** — "End-to-end migrator for AWS data stack → Oracle AIDP".
- **`README.md:95`** — a section titled **"Real AWS (live mode)"** whose own example command is
  `aws-aidp migrate plan.json --filter athena --demo`. The title promises live; the command three
  lines down is offline. "Live mode" is doing double duty here for *live AWS reads*, which work,
  and *live AIDP writes*, which do not — and the docs never separate the two.
- **`README.md:209`** — "v0.5 — live AIDP write-side end-to-end (currently demo-only)". The
  accurate statement exists, at the bottom of the roadmap, after two hundred lines presenting a
  migrator.

Fixing these line by line leaves the frame intact, and the next revision reintroduces the same
class — because the name and the opening still invite it.

## What it actually is, and why that's worth saying plainly

An inventory tool and a deterministic translator that emits reviewable artifacts and flags what it
cannot convert safely. That is genuinely useful on its own: translating an Athena and Glue estate
is exactly the tedious work nobody wants to do by hand, and *flag rather than guess* is the right
call — the "What PASS means" note is one of the more honest things in this repo.

None of that needs a migration claim to justify it.

## Suggested repositioning

Cheapest version, text only:

| Where | From | To |
|---|---|---|
| Opening line | "End-to-end migrator" | something like "Migration assistant … inventories an AWS data stack and translates Athena SQL and Glue ETL to Spark on AIDP. Applying the output is manual today" |
| `## Real AWS (live mode)` | | `## Real AWS (live inventory)` |
| Roadmap line on v0.5 | line 209 | move it into Status, or into the opening |
| Status table | `migrate` ✅ | note that it writes artifacts locally; nothing is applied to AIDP |
| `.[oci]` comment | "for live AIDP calls" | drop it until a verb reaches the client |

Whether the plugin should also be renamed — `aws-aidp-migrator-assist`, or similar — is a
maintainer call, and the marketplace entry name is already in flux per the review, so if it's going
to happen this is the cheap moment. I'd weight the README framing well above the rename: the name
is what people see once, the opening paragraph is what they act on.

## One scope question for the maintainers, not the author

Version is 0.2.0 and the roadmap puts EMR, SageMaker and the live write-side at v0.3 through v0.5.
That's fine for a tool; it's a question for a samples catalog. Does a sample carry a roadmap of
unimplemented verbs, or does it ship the slice that works and grow later? That's the same question
the `capability-probes` folder raises in #107 from the other direction, and it would be worth
settling once rather than per PR.

To be clear about what this is: not a defect and not a blocker from my side. The engineering here
is careful and I'd want it in the catalog. I'd just want the first paragraph to promise what the
fourth verb delivers.

ahmedawan-oracle:
@mppotnuru I ran a Claude Code plugin-standards validation pass on this branch (strict YAML parse, `claude plugin validate`, live checks) and opened a fix PR against your `add-aws-migrator-claude` branch — merging it updates this PR directly:

**https://github.com/mppotnuru/oracle-aidp-samples/pull/1**

It fixes the four standards blockers, two of which were silently disabling the plugin:

1. `commands/inventory.md` — `argument-hint` starting with `[` is invalid YAML (flow-sequence parse error), so Claude Code dropped the inventory command entirely. Quoted.
2. `skills/aws-aidp-migrator/SKILL.md` — `(verbs: inventory, ...)` in the unquoted description is a YAML parse error, so the skill never registered. Quoted. (This is the frontmatter P1 from the earlier review — confirmed by parser; plan/migrate/verify were already valid.)
3. `marketplace.json` entry name now matches `plugin.json` (`oracle-ai-data-platform-workbench-aws-migrator`), matching all merged sibling plugins; marketplace renamed `aidp-aws-migrator` in sibling style.
4. README install flow — the repo has no root `.claude-plugin/marketplace.json`, so `/plugin marketplace add oracle-samples/oracle-aidp-samples` can't resolve; now documents adding the marketplace via the plugin directory, installing as `name@marketplace`, the correct slash-command namespace, and drops the nonexistent `cd aws-aidp-migrator` step.

Deliberately **not** included (still open from the review threads): `docs/MCP.md`/`DEMO.md` broken links, the `live_seed.py`/`live_teardown.py` references + the test that reads them, `.mcp.json` `python` → `python3` and the `mcp` optional-dep startup failure, `pip install aws-aidp-migrator` (not on PyPI — returns 404), the nested `.github/workflows` file (GitHub only runs root-level workflows), and the unrecognized `displayName`/`skills` fields in `plugin.json`.

mppotnuru:
Round-2 items are addressed and pushed — this branch now carries the full set.

**Athena — the four new PASS-then-wrong classes, plus the two that had not landed**

- `date_format` literal escaping: the doubled quote is gone. It now emits `'yyyy-MM-dd\'T\'HH:mm:ss\'Z\''`, and that output was executed on Spark 3.5.9 and checked to return `2024-01-03T10:20:30Z` rather than string-matched. The corpus test that asserted the broken string as expected is replaced, and there is now an invariant test that no `date_format` rewrite can emit `''` again.
- `string_literal_escaping`: `'O''Brien'` is flagged. Same root cause as the above — Spark closes the literal and concatenates, so it was silently matching `OBrien`.
- `array_constructor`: `ARRAY[...]`, `MAP(ARRAY..., ARRAY...)`, `array(varchar)` and `row(...)`, one finding per construct.
- `integer_division`: `7/2` flagged. Deliberately conservative — integer literals and provably-integral functions, not a bare `a/b`, since operand types need a catalog this tool does not have.
- `day_of_week`: `EXTRACT(DOW FROM x)` now flagged alongside the function-call spelling patched in round 1.
- `greatest`/`least` NULL divergence flagged.

**Glue — the GlueContext-method gap**

The residual-API check listed three names, so sixteen GlueContext-only methods reached PASS and then raise `AttributeError` on a SparkSession: `forEachBatch`, `purge_table`, `purge_s3_path`, `transition_table`, `transition_s3_path`, `start_transaction`, `commit_transaction`, `cancel_transaction`, `write_from_options`, `create_data_frame_from_catalog` / `_from_options`, `extract_jdbc_conf`, `add_ingestion_time_columns`, `getSourceWithFormat`, `getSinkWithFormat`, `getSampleStreamingDynamicFrame`.

Those skew heavily toward the destructive jobs — purge, transition, transaction — which is the worst place to be quiet. Verified against the masked output the rule actually sees, with a control job that stays clean so the check is not simply flagging every Glue job.

**Plugin standards**

The unparseable YAML frontmatter is fixed in `commands/inventory.md` and `SKILL.md`. You were right that the skill was loading with empty metadata, and that it predated round 1. Marketplace/plugin name alignment, the install flow, and the stale CHANGELOG/TESTING/README are updated; version is 0.3.0. `test_plugin_frontmatter.py` now guards the frontmatter so it cannot regress silently.

**The two red tests**

`live_teardown.py` is development-only and excluded from the published tree, so that test now skips when the script is absent rather than failing in a tree where its subject was never shipped. The scale test's 10s bound was a hardware benchmark rather than a correctness check — cost is linear in input size (measured 0.69 / 1.35 / 2.71 / 5.50s at 10k / 20k / 40k / 80k lines), so it is now a 60s hang ceiling with those measurements recorded beside the constant.

369 tests pass, and I ran the suite inside both regenerated plugin trees rather than only at source. That is how one more of the same class turned up: the new frontmatter test asserted `commands/*.md` exist, which fails in the Codex tree because slash commands are a Claude Code concept and that tree drives the verbs over MCP. It now requires them only where the directory exists.

Not claimed as done: no translated Glue job has been executed on a live cluster, so the widgets mapping stays flagged rather than promoted to a rewrite.

ahmedawan-oracle:
@mppotnuru Round-2 plugin-standards validation on the updated head (`621f6c9`) — same method as before: strict YAML parse, `claude plugin validate`, live checks, and a full local test run. Great progress: **3 of the 4 prior blockers are verifiably fixed**, and the suite is green on my machine (pytest: 377 passed / 25 skipped / 548 subtests; stress runner: 369 tests OK). The frontmatter regression guard in `tests/stress/test_plugin_frontmatter.py` is real — it lints exactly the failure classes that shipped. Security sweep stays clean.

Two blockers remain, both one-line doc fixes:

1. **The documented install still cannot resolve.** `README.md:47` says `/plugin marketplace add oracle-samples/oracle-aidp-samples`, but the repo root has no `.claude-plugin/marketplace.json` (verified: `claude plugin validate` on the repo root reports "No manifest found"). The only root manifest is the Codex-format `.agents/plugins/marketplace.json`, which doesn't list this plugin. Suggest documenting the community-marketplace path the way the merged databricks-migrator README does, or (maintainer call) adding a real root-level Claude Code marketplace manifest for the whole repo.
2. **`SKILL.md:19` still offers `pip install aws-aidp-migrator`** — the package is not on PyPI (404). Because Claude executes skill instructions, an unregistered package name in a published Oracle sample is a name-squatting supply-chain vector, not just a broken doc. Please drop the fallback (keep `pip install -e .`) or reserve the name before merge.

Worth folding into the same commit:

- `.claude-plugin/plugin.json` — remove the unrecognized `displayName` and `skills` fields (tolerated but inert; `commands`/`mcpServers` are fine).
- `.mcp.json:4` — `"command": "python"` → `"python3"` (clean macOS/Linux installs have no `python`; every other invocation in the plugin already uses `python3`). Also consider a README note that the MCP server needs `pip install -e '.[mcp]'` (Python 3.10+) or it will show as failed in `/mcp` on a fresh install — the SystemExit message you added is good, but users see the failure before they see the message.
- `README.md:58` — "Codex users can also install this plugin from the Codex marketplace in this repository" isn't true at this head: the Codex marketplace lists 5 plugins and this isn't among them. Remove the sentence or add the Codex entry in this PR (or note it lands via #109).
- The nested marketplace is now named `oracle-aidp-samples` — a single-plugin marketplace four directories deep named as if it were the whole repo will collide if a genuine root marketplace ever lands. Suggest a plugin-scoped name.

Nits, take or leave: `--filter` hints in `commands/migrate.md`/`verify.md` and `references/verbs.md:25` say `athena|glue` while `cli.py` accepts all five sources; `CHANGELOG.md:32` says 343 tests vs 369 at head; `SKILL.md:58` points at `README.md` which doesn't exist relative to the skill directory.

Fix those two blockers and, from the Claude Code plugin-standards side, this is good to go.

mppotnuru:
Both blockers fixed, plus everything in the fold-in and nit lists. Pushed.

**Install path.** You were right that it could not resolve — no root `.claude-plugin/marketplace.json` exists. The README now documents Anthropic's community marketplace as the primary route, following the merged databricks-migrator README (including the weekly-cadence caveat), with the local-clone path kept for pre-merge use.

**PyPI fallback.** Removed. Agreed this is a security issue rather than a broken link — an agent executes skill instructions verbatim, so an unregistered package name is a name-squatting vector. I left a one-line note saying why it is absent, so it does not get helpfully restored later. `pip install -e .` and a `${CLAUDE_PLUGIN_ROOT}` form remain.

**Folded in**

- `plugin.json`: dropped the inert `displayName` and `skills` keys.
- `.mcp.json`: `python` → `python3`. The Codex variant is generated by `build_pr_trees.sh` and had the same bug, so it is fixed at the generator rather than only in the checked-in file — otherwise #109 would still have shipped it.
- README now states the `pip install -e '.[mcp]'` (3.10+) requirement up front, since the `/mcp` failure is what users see first.
- Removed the Codex-marketplace sentence; it now points at #109 instead.
- Nested marketplace renamed to `aidp-aws-migrator`. Your original naming was right and I had wrongly kept `oracle-aidp-samples` when resolving the conflicts on your fix PR — a single-plugin marketplace four directories deep should not be named as if it were the whole repo.

**Nits**

- `--filter` hints in `commands/migrate.md`, `commands/verify.md` and `references/verbs.md` now list `s3|glue|athena|emr|sagemaker`, matching `_FILTER_TYPES`.
- CHANGELOG 343 → 369.
- `SKILL.md` now links `${CLAUDE_PLUGIN_ROOT}/README.md`, which resolves from the skill directory.

Verified at this head: 369 stress tests OK and the legacy suites pass in the source tree and in both regenerated plugin trees; every frontmatter block parses with a strict YAML parser; all four manifests are valid JSON; and the marketplace name now matches the documented install command.

Thanks for the three rounds — the frontmatter and PyPI findings in particular were things no amount of reading the diff would have caught.

ahmedawan-oracle:
@mppotnuru Independently re-verified at head `0b5ba8a` before signing off — same method as the prior rounds, not a read of your comment:

- All five frontmatter blocks strict-parse (PyYAML), descriptions present
- All four manifests are valid JSON; marketplace entry name == `plugin.json` name; `aidp-aws-migrator` naming matches the documented `name@marketplace` install line; inert `displayName`/`skills` keys gone; `.mcp.json` uses `python3`
- `claude plugin validate` passes
- Full local suite green on Python 3.14: 377 passed / 25 skipped / 548 subtests (stress runner: 369 OK)
- README install paths, the `` pip form, the MCP prerequisite framing, the #109 Codex pointer, and all three nits confirmed in place

**No remaining Claude Code plugin-standards findings — the standards review is closed from my side.** Fixing the `.mcp.json` bug at the `build_pr_trees.sh` generator rather than the checked-in file was the right call, for the same reason as round 1.

Remaining merge gates are outside plugin standards: @navid-heydari's re-review of the translator P1s this head claims to address (date_format escaping, Presto constructors, integer division, EXTRACT(DOW), greatest/least NULLs, GlueContext-only methods), and the README framing question from @lucas-de-almeida-reis, which is a maintainer positioning call rather than a defect.

mppotnuru:
@lucas-de-almeida-reis — applied the text-only repositioning. Your diagnosis was the useful part: I had been fixing those lines individually across three rounds, and they kept coming back because the opening paragraph invited them.

| Where | Now |
|---|---|
| Opening line | A migration assistant that inventories an AWS stack and translates Athena SQL and Glue ETL to Spark, stating in the first paragraph that applying the output is manual today |
| `## Real AWS (live mode)` | `## Real AWS (live inventory)` |
| Status, `migrate` row | Says artifacts are written locally and nothing is applied to AIDP |
| `.[oci]` comment | No longer claims to be "for live AIDP calls" |
| Roadmap v0.5 | Points at Status rather than carrying the only accurate statement on its own |

On the Glue Data Catalog: you were right, and it was worse than a wording slip. I ran the offline pipeline and checked the per-kind breakdown — all 50 `aidp_dcat_external_table` and 2 `aidp_schema` assets come back `planned`, because no DDL is emitted at all. `SKILL.md` was claiming "S3, Glue Data Catalog, Athena SQL, Glue ETL — working" as a single line. Coverage is now split per source, with the catalog listed as inventoried and planned but not translated, and the `rclone` job described as something you review before running rather than something the tool executes.

I did not rename the plugin. Taking your own weighting there — the name is seen once, the opening paragraph is acted on — and it is a maintainer call rather than mine to make inside this PR.

Your scope question about whether a sample carries a roadmap of unimplemented verbs is a fair one and I do not think it should be settled here. If the catalog convention ends up being "ship the slice that works", the EMR/SageMaker stubs and the v0.3–v0.5 roadmap are the parts that would come out, and I am happy to do that in a follow-up once it is decided alongside #107.

Thanks — this was the most useful kind of review to get, because it was about the thing generating the defects rather than the defects.

ahmedawan-oracle (approved):
Approving after three verification rounds (see comments): plugin-standards findings all fixed and independently verified at each head — frontmatter strict-parses, manifests valid and name-aligned, `claude plugin validate` passes, full suite green locally (377 passed / 25 skipped / 548 subtests), security sweep clean. Translator P1 fixes are covered by the passing semantic-gap / GlueContext / output-validation test gates. README repositioned per @lucas-de-almeida-reis. I refreshed the PR description at merge time to match (the prior body predated the repositioning). @navid-heydari — merging as maintainer with the P1s test-verified; happy to address anything further in a follow-up.

PR #109, opened 2026-09-11
Add AWS → Oracle AIDP Migrator (Codex plugin)

mppotnuru:
New Codex plugin under `ai/codex-plugins/` — same AWS→AIDP migrator, exposed to Codex / Cursor / Claude Desktop via an **MCP server** (`aws-aidp-mcp`; tools: inventory / plan / migrate / verify).

- `.codex-plugin/plugin.json` + skills + README + MIT.
- Deterministic core (sqlglot + rclone), same tested engine as the Claude plugin (separate PR).
- MCP server requires Python 3.10+.


ahmedawan-oracle:
## Deep review — same engine as #108 (every engine finding there applies here byte-for-byte); Codex-specific: the plugin sits outside `plugins/` and is unregistered so `codex plugin add` cannot resolve it, the MCP server — its only integration surface — dies on a fresh install and hangs on Windows, and the docs are the Claude-oriented personal-repo docs with broken links

@mppotnuru — reviewed by execution against the repo's Codex conventions (both marketplace manifests, `ai/codex-plugins/README.md` + `TESTING.md`, all five sibling `.codex-plugin/plugin.json` files and their 98 SKILL.md frontmatters), the documented install path (`pip install -e '.[mcp]'`), the MCP server driven over stdio with a real `mcp` client, a wheel build + non-editable install, and `diff -rq` against #108. The `aws_aidp/`, `tests/`, `scripts/`, `fixtures/`, README, TESTING, pyproject, `.mcp.json`, LICENSE, CHANGELOG and `skills/` are byte-identical to #108, and `MCP.md` here is identical to #108's `docs/MCP.md` — so the translator, planner, verify, packaging, live-path and content findings in my #108 review (https://github.com/oracle-samples/oracle-aidp-samples/pull/108#issuecomment-5645630597) all apply to this PR and are not repeated in full below.

### Verified working ✔

- **Parity holds**: the "same tested engine" claim in the body is accurate (`diff -rq` differs only in `.claude-plugin/`, `commands/`, `docs/` vs `.codex-plugin/`, `MCP.md`); 19/19 tests pass from this tree; the offline fixture pipeline runs (113 assets planned; verify PASS 20 / REVIEW 13 / SKIP 80 / FAIL 0).
- **The MCP server design is sound once two fixes land**: with `mcp` pinned below 2 and stdin detached (see below) it initialises (`serverInfo` aws-aidp-migrator, protocol 2025-11-25), exposes exactly the four documented tools with sane schemas (`plan.manifest` and `migrate.plan_path` required; `migrate.demo` defaults true — a safe offline default), and the full `inventory(fixture="demo") → plan → migrate → verify` chain ran end-to-end over MCP, all `isError=False`. The `MCP.md` tool table matches `list_tools()` exactly, and its `[mcp_servers.aws-aidp]` `config.toml` snippet matches the documented Codex MCP config shape.
- **Manifest basics match siblings**: `.codex-plugin/plugin.json` is valid JSON, `"mcpServers": "./.mcp.json"` and `"skills": "./skills/"` follow ask-aidp / fusion-autopilot; the SKILL.md frontmatter uses exactly the `name`/`description` fields every sibling skill uses; `references/verbs.md` resolves.
- **Secret hygiene**: only `xxxx` placeholders and the synthetic Acme account id; no e-mails or hostnames; LICENSE is standard MIT with the Oracle copyright line the siblings use.

### Blockers

1. **Wrong location, no registration — `codex plugin add oracle-ai-data-platform-workbench-aws-migrator@oracle-aidp-codex` has nothing to resolve.** All 55 files live under `ai/codex-plugins/oracle-ai-data-platform-workbench-aws-migrator/`; every existing Codex plugin lives under `ai/codex-plugins/plugins/<name>/` and is listed in **both** `.agents/plugins/marketplace.json` (`./ai/codex-plugins/plugins/<name>`) and `ai/codex-plugins/.agents/plugins/marketplace.json` (`./plugins/<name>`) with a `policy` block and `category`. `git diff 88ebf4b..42f6078 -- .agents ai/codex-plugins/.agents ai/codex-plugins/README.md ai/codex-plugins/TESTING.md` is empty — neither manifest, the plugin table, the `codex plugin add` list nor the layout tree knows this plugin. `ai/codex-plugins/TESTING.md` documents that `source.path` resolves relative to the marketplace root, so the only documented Codex install path cannot work (the codex CLI is not installed on my machine, so this is established from the manifests, not executed). Fix: move to `ai/codex-plugins/plugins/<name>/`, add entries to both manifests (`policy: {installation: AVAILABLE, authentication: ON_INSTALL}`, `category: Engineering`), add the README row + install line + layout entry and a TESTING.md section.
2. **The MCP server cannot start after the documented install.** `MCP.md`/README say `pip install -e '.[mcp]'`; `pyproject` declares `mcp>=1.2` unbounded, which today resolves to **mcp 2.2.0** where `mcp.server.fastmcp` was removed (renamed `MCPServer`). `aws_aidp/mcp_server.py:18` fails and the `except ImportError` masks the real error as `The MCP SDK is required. Install it with: pip install -e '.[mcp]'`. Executed: `aws-aidp-mcp` and `python -m aws_aidp.mcp_server` both exit 1 (`pip_install.log`: `Successfully installed … mcp-2.2.0`). For #109 this is the *only* Codex integration surface, so the plugin is dead on arrival for any fresh install. Fix: `mcp = ["mcp>=1.2,<2; python_version >= '3.10'"]` (or port to `mcp.server.mcpserver.MCPServer`), print the real ImportError, add a smoke test.
3. **`.mcp.json` spawns a bare `aws-aidp-mcp` that nothing installs for a Codex user.** The console script exists only after a pip install from the plugin directory; no skill, hook or README step performs it (README's only install text is `git clone <this-repo>` + `pip install -e .`, and the Codex section links a file that doesn't exist — below). Siblings avoid this: ask-aidp uses `"cwd": ".", "command": "node", "args": ["./mcp/…"]`; engineer-agent and databricks-migrator stage their Python engine from a `hooks/session_start.py`. Fix: `"command": "python", "args": ["-m", "aws_aidp.mcp_server"], "cwd": "."` plus a bootstrap skill or SessionStart hook that runs `pip install -e "<plugin root>[mcp]"` once (or `uvx --from <path> aws-aidp-mcp`), and Codex install steps in the README.

### Majors — Codex-specific (executed)

- **Every MCP tool call hangs on Windows.** With `mcp` pinned to 1.30 so the server starts, `list_tools` succeeds but `inventory(fixture="demo")` never returns: the spawned `python -m aws_aidp.cli …` child sat 10+ minutes at 0.03 s CPU / 1 thread, blocked at interpreter start, while the identical command run directly finishes in 2.4 s. Root cause: `_run` (`mcp_server.py:29`) uses `subprocess.run(capture_output=True)` and the child inherits the MCP transport's stdin pipe. Controlled experiment with my own FastMCP server: the plugin's exact call pattern blocked until a 90 s timeout killed it (`rc=143`); the same call with `stdin=subprocess.DEVNULL` returned in 0.2 s. With that one-line patch on a copy, all four tools completed. Reproduced via both the pip launcher and `python -m`; macOS/Linux not tested. Better still: call `aws_aidp.cli.main(args)` in-process with redirected stdout/stderr — no child, real exit codes.
- **README links three files that don't exist in this tree.** `README.md:57` sends Codex/Cursor users to `[docs/MCP.md](docs/MCP.md)` — `MCP.md` was relocated to the plugin root and `docs/` doesn't exist here (`scripts/build_pr_trees.sh:76-77` copies README verbatim and moves the file without rewriting the link); `README.md:66` `cp .env.example .env` (only `demo.env.example` ships, with different variables); `README.md:145` lists `DEMO.md`. The rest of README/TESTING is the personal repo's Claude-Code-oriented text: `git clone <this-repo>`, `/plugin marketplace add mppotnuru/aws-aidp-migrator` (a **private** repository per `gh api`), `/aws-aidp-migrator:inventory` slash commands that do not exist in Codex, `TESTING.md:14` cloning the personal repo, `TESTING.md:20` "the only runtime dependency is boto3" (sqlglot is a hard import). Nowhere does the plugin explain `codex plugin add …@oracle-aidp-codex`, how the `aws-aidp-migrator` skill is invoked, or that the MCP server is auto-registered by `plugin.json`. Write a Codex-specific README.
- **`MCP.md:3-4` claims "OpenAI's Codex CLI has no plugin marketplace"** — contradicted by the repository this PR targets (`codex plugin marketplace add oracle-samples/oracle-aidp-samples`, verified there on codex-cli 0.141.0) and by the PR's own `.codex-plugin/plugin.json`. `MCP.md:38-40` then tells Codex users to hand-edit `~/.codex/config.toml` `[mcp_servers.aws-aidp]`, which would register a second `aws-aidp` server next to the one the plugin manifest already declares. Keep the manual snippet for non-plugin clients (Cursor, Claude Desktop) only.
- **`plugin.json` diverges from all five sibling manifests**: every sibling carries an `interface` block (`displayName`, `shortDescription`, `longDescription`, `developerName`, `category`, `capabilities`, `defaultPrompt`, mostly `websiteURL`/`privacyPolicyURL`/`termsOfServiceURL`/`brandColor #C74634`) and four carry `homepage` + `repository`. This one has a top-level `displayName` (the Claude Code field), no `interface`, no `homepage`/`repository`, author "Oracle AI Data Platform", and ships no `PRIVACY.md`/`NOTICE` although `ai/codex-plugins/README.md:87` says each plugin has license/privacy files (4 of 5 siblings do). The OpenAI manifest reference could not be fetched in usable form, so this is rated against the in-repo convention.
- **Shared engine blockers/majors** (detail and evidence in the #108 comment): translated Glue jobs call `oidlUtils.widgets.get`, an API the live-validated AIDP toolkit does not use for Python (never executed live); the Athena transpiler silently drops `array_agg ORDER BY`, mistranslates sub-day `date_add`, integer division, `day_of_week`, `to_hex(sha256)`, `log(b,x)`, `concat_ws` NULLs, discards sqlglot Unsupported warnings, truncates multi-statement input and passes unknown functions through as PASS (demo q-1004 and q-1015 are wrong/failing PASSes on Spark 3.5.9); the Glue translator forces `.mode("overwrite")` and drops `partitionKeys`; the planner emits a malformed `oci:@ns//bucket/…` for all 50 tables; `<your-oci-namespace>` ships inside PASS artifacts; `migrate` crashes with `UnicodeEncodeError` on Windows default encoding; `fixtures/` is not packaged (non-editable install breaks `--fixture demo` and the MCP `inventory(fixture="demo")` tool — executed here too, exit 2); `pyproject` says Apache-2.0 while LICENSE/plugin.json say MIT (wheel METADATA carries both); `live_teardown.py` deletes by fixed name with no ownership check; the "verified end-to-end on live AIDP" claim is a hand-written smoke script that never consumes migrator output.

### Minors (executed)

- MCP tools report CLI failures as ordinary text with `isError=False`, stderr merged into stdout and full Python tracebacks with absolute local paths (`plan(manifest='nope.json')` → 1,488-char traceback); a Codex agent cannot structurally tell success from failure. Raise `ToolError`/set `isError` on non-zero exit and validate paths first.
- Relative defaults (`inv.json`, `plan.json`, `migrated`) resolve against the **server** process cwd, which the Codex user doesn't control unless they add `cwd` to config.toml: executed with distinct server/client cwds, `inv.json` landed in the server cwd. Require absolute paths or honour MCP `roots`.
- `migrate(demo=False)` is a silent no-op that exits 0 with `error=0` and writes a report (`runner.py:132-142` marks Athena/Glue `planned: live mode not yet wired`), while `runner.py:7` says live mode "calls AIDP REST via aidp_client/" and the MCP schema exposes `demo` as a real toggle. Fail fast or drop the parameter until it does something.
- `fixture` is interpolated into a path unsanitised (`inventory(fixture='../../etc')` → traversal honoured, exit 2); validate against the bundled list.
- The live scripts hard-depend on the **Claude** plugin cache path for the sibling's `aidp_sql.py` (`demo.env.example:24`, pinned to engineer-agent 0.5.0); the Codex engineer-agent sibling ships that helper at `aidp/aidp_sql.py`, so a Codex-only user cannot run the PR's "live" path at all.
- `CHANGELOG.md` describes the Claude tree (`.claude-plugin/`, `commands/*.md`, `docs/FDE_PRESENTATION.md`, `docs/MCP.md`) — none exist here — and says 15 tests (19); `aws-aidp --version` prints 0.1.0 (pyproject/plugin.json 0.2.0).
- **5.5K lines duplicated byte-for-byte across the two PRs with no canonical source**: the repo's convention for shared content is `ai/shared-plugin-content/<plugin>/` + a sync script that copies and `filecmp`-verifies into both trees (spark-connectors); the only "sync" here is `scripts/build_pr_trees.sh`, which lives outside the repo workflow, `rm -rf`s a path under `~/Documents`, needs `docs/` and `.claude-plugin/` (absent from this tree, so it cannot even run from here), and has already produced drift (the `docs/MCP.md` link). Put the engine under `shared-plugin-content` with a sync script and keep the two plugin dirs as thin manifests + variant READMEs; drop `build_pr_trees.sh` from the shipped plugin.

### Note

Merge gate you already know about: OCA unsigned (CI red), no `Signed-off-by`, PR template/issue not used. The 3.7 MB of demo media is (rightly) absent from this tree.

**Bottom line:** the Codex variant is one marketplace entry, one `mcp<2` pin and one `stdin=DEVNULL` away from working — but today it cannot be installed, cannot start, and would hang on Windows if it did; fix those three, write a Codex README, bring `plugin.json` in line with the sibling manifests, then pick up the shared-engine items from the #108 review. Happy to re-verify. 🤖 Deep review by Claude (Fable)

mppotnuru:
Thanks for the exceptionally thorough review — force-pushed a substantial revision that addresses it.

**Engine (both PRs):** the Athena transpiler was reworked off `sqlglot` onto a hardened deterministic translator with explicit semantic-gap guards, so the silent-semantics findings (`array_agg ORDER BY`, sub-day `date_add`, integer `/`, `day_of_week`, `to_hex(sha256)`, `log` arity, `concat_ws` NULLs, unknown-function passthrough, multi-statement truncation, sqlglot version drift) are gone — those cases now flag rather than PASS, backed by a corpus + Spark-runtime tests. Glue now emits `oidlUtils.parameters.getParameter(...)`, flags method-form DynamicFrame APIs, and `verify` PASS requires a compile + unbound-name scan.

**Fixed here:**
- MCP: `mcp>=1.2,<2` (+ py>=3.10 marker); `stdin=DEVNULL` (Windows hang); self-installing `.mcp.json` (`python -m` + `${CLAUDE_PLUGIN_ROOT}` / cwd for Codex); real ImportError surfaced.
- Packaging: `fixtures/` now under `aws_aidp.fixtures` (packaged); planner OCI location bug fixed; Windows `UnicodeEncodeError` fixed (utf-8); collision-safe artifact paths; namespace validated.
- Identity: plugin name/namespace/URLs point at oracle-samples; private-repo + dead slash-command refs removed; broken links fixed.
- Licensing/hygiene: MIT aligned across pyproject/LICENSE/plugin.json; added `NOTICE` + `PRIVACY.md`; rclone inline-comment fixed; unsourced market stat + non-AIDP "DCAT" vocabulary removed; presenter kit, live scripts, build tooling and demo media removed from the shipped tree; no internal names/OCIDs.
- Process: `Signed-off-by` on all commits (OCA green); registered in the plugin catalog / marketplace manifests.

Remaining follow-ups called out honestly: one live run of a translated Glue job on a cluster to promote the widgets mapping from flagged→rewrite, and the shared-plugin-content dedup between the two trees. Happy to iterate. 🙏

**Codex-specific:** moved under `ai/codex-plugins/plugins/`, registered in both `.agents/plugins/marketplace.json` manifests + the README table; added a Codex-specific README (no Claude slash-command text); `.mcp.json` launches `python -m aws_aidp.mcp_server` with `cwd: "."`; MCP tool calls no longer hang on Windows (`stdin=DEVNULL`).

mppotnuru:
Update (additive commit, no force-push): the **"verify PASS is a status relabel"** major is now addressed.

Root cause was exactly as you found — when the SQL side moved off `sqlglot` to regex rules, it lost the free parse-validation the transpiler gave for granted (`parse_failed` finding count went 1 → 0). Restored it as explicit **output-validation gates** in `athena_to_spark_sql.translate()`, mirroring the `ast`-parse gate the Glue translator already had:

- **multi-statement**, **unrecognised/unbalanced statement**, and **10 Presto-only constructs** Spark rejects (`WITH RECURSIVE`, `FETCH FIRST`, `CAST AS JSON`, CTAS `WITH(...)`, `> ANY(...)`, `AT TIME ZONE`, `TABLESAMPLE`, `PREPARE/EXECUTE`, `UNLOAD`) → now flag → REVIEW instead of PASS.
- A **Spark-3.5 builtins allowlist** so unknown functions are caught rather than passed through.
- Verdict wording made honest across report.html / verbs.md / README: **PASS = "translated, no known issue detected — not execution-verified."**
- Adversarial 20-query junk manifest that previously returned PASS 20/20 (incl. `CREATE EXTERNAL TABLE … s3://…` byte-identical) now flags correctly; added `test_output_validation.py` + `test_verdict_claims.py`.

Also included the earlier Spark-3.5.3 translator fixes (JOIN-after-UNNEST, 1-based subscripts, scoped `.spark_session`). Full stress suite green. Thanks again — happy to re-verify.

ahmedawan-oracle:
## Re-review of `aa46d16` — the three Codex blockers are fixed and verified (location + registration, MCP starts, no Windows hang); what remains here is the unchanged `plugin.json`, a few Codex-doc contradictions, whitespace churn in two shared manifests, and the shared-engine items in the #108 review

@mppotnuru — verified by execution on this head (engine tree byte-identical to #108's `d962dff`, so the translator, verify, packaging and live-path findings in my #108 re-review (https://github.com/oracle-samples/oracle-aidp-samples/pull/108#issuecomment-5668605532) apply here unchanged and are not repeated). Suite from this tree: 359–360 passed, 2 failed (the deleted-`live_teardown.py` test and the `test_scale` timing bound — see #108), 8 skipped.

### Verified fixed ✔

- **Location and registration**: the plugin now lives at `ai/codex-plugins/plugins/oracle-ai-data-platform-workbench-aws-migrator/`, appears in **both** `.agents/plugins/marketplace.json` files with a `policy` block and `category: Engineering`, path style matches the siblings, and the `ai/codex-plugins/README.md` table row (version 0.2.0 == `plugin.json` == `aws-aidp --version`) follows the sibling format.
- **MCP server starts and works**: `mcp>=1.2,<2; python_version >= '3.10'` resolves to mcp 1.30; `initialize` + `list_tools` (4 tools) + every tool call return in 0.2–0.5 s on Windows (`stdin=DEVNULL`), full `inventory(fixture="demo") → plan → migrate → verify` chain over MCP = PASS 13 / REVIEW 20 / SKIP 80 / FAIL 0; fixture mode works without boto3; `migrate(demo=False)` now returns `[exit 2] … live AIDP migration is not implemented` instead of a silent success.
- **`.mcp.json` `cwd: "."`** matches the sibling ask-aidp convention and resolves to the plugin root under Codex; the server starts from there.
- **Codex-specific README** written (marketplace install, `pip install -e "<plugin dir>[mcp]"`, tool table, "you do not need to hand-edit config.toml"); `TESTING.md`/`MCP.md` present; no binaries, presenter kit, live scripts or internal names in the tree; `Signed-off-by` on both commits.

### Codex-specific findings (executed)

- **[NOT FIXED] `.codex-plugin/plugin.json` is byte-identical to the round-1 head.** Its description still says "deterministic-first **via sqlglot** + rclone" (sqlglot is gone from `pyproject` and the code), and it still lacks the `interface` block (`displayName`/`shortDescription`/`longDescription`/`developerName`/`category`/`capabilities`/`defaultPrompt`/`brandColor`), `homepage`, `repository` and `privacyPolicyURL` that every sibling Codex manifest carries — `PRIVACY.md` now exists but nothing points at it; the top-level `displayName` is the Claude Code field. Mirror ask-aidp's manifest.
- **[NEW] Default outputs land inside the installed plugin directory.** With `cwd: "."` and the relative tool defaults (`inv.json`, `plan.json`, `migrated`), a simulated Codex session left `inv.json`, `plan.json` and `migrated/` in the plugin root — under Codex that is `~/.codex/plugins/<plugin>/`, which the next marketplace upgrade overwrites and the user never looks in. Default to an explicit user location (or require absolute paths) and say where relative paths resolve in `MCP.md`.
- **[NEW] Both shared marketplace manifests were re-serialised from 2-space to 1-space indentation** — `git diff --stat origin/main` shows 142 changed lines per file for a 12-line addition (`--ignore-all-space` collapses each to exactly the new entry). Re-emit with `indent=2` so the diff is only your entry; as shipped it will conflict with any concurrent plugin PR and hides the real change.
- **Codex docs contradict each other and the repo**: `README.md:3` and the catalog row scope the plugin to "(S3, Glue, Athena)" while `plugin.json`, `SKILL.md` and `pyproject` say "(S3, Glue, Athena, EMR, SageMaker)"; `README.md:10` `codex plugin marketplace add oracle-samples/oracle-aidp-samples` omits the `--ref main --sparse .agents --sparse ai/codex-plugins` flags that `ai/codex-plugins/README.md` and `TESTING.md` document as the working form; `MCP.md:3` still says "OpenAI's Codex CLI has **no plugin marketplace**" (this PR registers in one) and `MCP.md:22-27` tells Codex users to hand-edit `~/.codex/config.toml` `[mcp_servers.aws-aidp]` while `README.md:17-18` says they must not — following both registers a second `aws-aidp` server. Scope `MCP.md`'s manual snippet to non-plugin clients (Cursor, Claude Desktop).
- **Shared with #108, called out because they bite Codex users directly**: the `SKILL.md` frontmatter does not parse as YAML (unquoted `description:` containing `(verbs: inventory, …)`), so the skill's description — its discovery trigger — is silently dropped at load time (verified with Claude Code's loader; the same file ships here); MCP tool failures are returned as plain text with `isError=False`, so a Codex agent cannot structurally tell `[exit 2]` from success; the `ImportError` cause is still masked; `TESTING.md` §6 tells testers to run the deleted `live_seed.py`/`live_teardown.py`, and `pytest -q` is red because a shipped test reads the deleted script; the `.github/workflows/stress-tests.yml` inside the plugin directory is inert (Actions reads only the repo root) while `TESTING.md` claims a CI matrix; `CHANGELOG.md` is untouched and describes the Claude tree and deleted files.
- **PR body not updated**: still "Deterministic core (sqlglot + rclone)". Please rewrite it to the regex rule set + gates and the offline-only scope.

**Bottom line:** the Codex packaging is now correct in the places that blocked it — it installs, registers, starts and runs. Left to do here: bring `plugin.json` in line with the siblings and drop "sqlglot", fix the output-location default and the manifest re-indentation, reconcile the three Codex docs, quote the SKILL frontmatter, and then take the shared-engine list from #108 (the string-literal, `ARRAY[]`, `date_format`-escaping regressions and the two unfixed drifts are the substantive ones). Happy to re-verify. 🤖 Deep review by Claude (Fable)

lucas-de-almeida-reis (commented):
No code findings from me here — the engine tree is byte-identical to #108 (verified by checksum on
the translator, CLI and runner), the Codex plugin index row is in place, and there are no
credentials or internal identifiers in the diff. Everything code-level is already covered by
@ahmedawan-oracle's two rounds, and per the convention established in the #109 review I won't
repeat the shared-engine items.

One framing note, following the same link-don't-repeat pattern. I left a comment on #108 about
positioning — that the doc inaccuracies already flagged share a single cause, the artifact
presenting itself as a migrator while what it implements is closer to a translator or assistant
for migration: https://github.com/oracle-samples/oracle-aidp-samples/pull/108#pullrequestreview-5223052755

That argument applies here, but **this README is already the more honest of the two** and it is
worth saying so. At 31 lines against 209 it drops the claims that caused the problem on the Claude
side: the source list is "S3, Glue, Athena" rather than five with EMR and SageMaker presented as
equals, there is no "Real AWS (live mode)" section, and the `migrate` row reads
**"translate (offline by default)"** — which is the accurate word.

Two residuals, both small:

- **`README.md:3`** — "Migrate an AWS data stack (S3, Glue, Athena) to Oracle AI Data Platform."
  Nothing is applied to AIDP; the output is artifacts to review and apply yourself.
- **"offline by default"** implies a configurable default with an online counterpart. There is one
  path, and non-demo `migrate` exits 2. "offline" alone would be accurate.

If the #108 framing lands, this file needs only those two touches. Not a blocker from my side.

mppotnuru:
Round-2 items are addressed and pushed — this branch carries the same engine set as #108, plus the Codex-specific fixes.

**Shared engine** (detail in the #108 comment): `date_format` now emits `'yyyy-MM-dd\'T\'HH:mm:ss\'Z\''`, verified by executing it on Spark 3.5.9 rather than string-matching; `string_literal_escaping`, `array_constructor`, `integer_division`, `EXTRACT(DOW)` and `greatest`/`least` NULL divergence are all flagged; and sixteen GlueContext-only methods (`purge_table`, `start_transaction`, `forEachBatch`, `transition_*`, …) no longer reach PASS before raising `AttributeError` on a SparkSession.

**Codex-specific**

- The unparseable `SKILL.md` frontmatter is fixed here too. Worth flagging explicitly: the fix PR against the Claude branch did not cover this tree, and since both trees ship the same `skills/` directory, the Codex skill was loading with empty metadata for exactly the same reason.
- `.codex-plugin/plugin.json` no longer references `sqlglot` — that dependency is gone from the code — and the version is now 0.3.0, matching `pyproject` and the CHANGELOG.
- `test_plugin_frontmatter.py` guards the frontmatter so it cannot regress silently.

One thing that only showed up because I ran the suite inside the regenerated Codex tree rather than at source: the new frontmatter test asserted the four `commands/*.md` exist, which fails here because slash commands are a Claude Code concept and this tree drives the verbs over MCP. It now requires them only where the directory exists. Same shape as the `live_teardown.py` failure — a test travelling to a tree where its subject was never shipped — so it seemed worth fixing the pattern rather than the one instance.

369 tests pass in this tree.

**Still open from your review, not claimed as fixed**

- `plugin.json` still has no `interface` block, so it remains out of line with the five sibling manifests.
- The default output paths still resolve inside the installed plugin directory, so a marketplace upgrade would wipe them.
- The marketplace manifest re-indentation noise is unchanged.

Happy to take those in a follow-up if you would rather keep this PR to the engine and frontmatter work.

ahmedawan-oracle (approved):
Approving: the three Codex-specific blockers were verified fixed by execution in my earlier re-review (location + registration, MCP startup, no Windows hang), the shared engine set matches #108 — which just merged after three verified validation rounds — and the author's round-2 Codex fixes are in. Merging as maintainer; anything residual can land as a follow-up.

PR #110, opened 2026-09-15
fix(medallion): correct type mismatch and gold schema drift in the telecom demo

navid-heydari:
# Description

`telecommunications_medallion_architecture_demo.ipynb` could not complete a Run All. The bronze write failed on a type mismatch between the schema `createDataFrame()` infers and the schema the DDL declares, and the fix that had been added for it took the form of extra cells *after* the failing ones — so the notebook shipped erroring by design. This corrects the cause instead and removes a schema-drift problem in the gold layer.

**1. Cast in the generating cell, not in a follow-up cell.**
`createDataFrame()` infers `double` for a Python float and `bigint` for an int, while the bronze DDL declares `DECIMAL(10,3)`, `DECIMAL(8,2)` and `INT`. A Delta `overwrite` through `saveAsTable` validates the DataFrame's schema against the table's and raises `DELTA_FAILED_TO_MERGE_FIELDS`, because decimal and double are not implicitly reconciled. The generating cell now casts each column to its declared type, and the three fix-up cells are gone.

**2. Replace the informal error markdown with the actual behaviour.** The three mismatch cases do not behave the same way, and the old note conflated them:

| Case | Result |
|---|---|
| Type mismatch | **fails** — `DELTA_FAILED_TO_MERGE_FIELDS` |
| Column in the table, absent from the DataFrame | **succeeds**, written as `NULL` |
| Column in the DataFrame, absent from the table | **fails** |

The second row is why adding `ingestion_timestamp` populates an audit field rather than avoiding the error; the third is why the silver step now ends with an explicit `.select(...)` of exactly the DDL columns, dropping the `ingestion_timestamp` it inherits from bronze.

**3. Remove `overwriteSchema` from the three gold writes.** It was silently replacing the declared `DECIMAL`/`INT` schema with the aggregates' widened types (`decimal(21,3)`, `double`, `bigint`) and dropping `created_at` entirely — so the gold DDL in the notebook did not describe the tables the notebook produced. Each aggregate is now `CAST` to its declared type, `created_at` is projected, and columns are in DDL order.

**4. Widen `total_call_minutes` to `DECIMAL(12,2)`** on `network_infrastructure` and `service_performance`. Summed across a service type it reaches roughly 1.2M minutes against a `DECIMAL(8,2)` ceiling of 999,999.99. `overwriteSchema` had been masking this, so removing it without widening would have silently produced `NULL`s.

**5. Drop the three gold tables before creating them.** A workspace that ran an earlier version of this notebook has those tables carrying the schema `overwriteSchema` inferred; `CREATE TABLE IF NOT EXISTS` would leave them as they are and the corrected writes would fail the schema check on exactly the workspaces the fix is for.

Net effect: 29 cells to 26, no cell relies on a later cell to repair it, and the declared DDL is authoritative in all three layers.

## Type of change

# How Has This Been Tested?

Verified statically, cell by cell:

- Every write's projection reconciled against its table's DDL — column names, order, and declared type vs. the type the expression produces.
- `DECIMAL` headroom computed from the generator's own parameters (10,000 subscribers × 20–100 events, per-service volume and duration ranges) for every summed column, which is how the `DECIMAL(8,2)` overflow was found.
- Notebook JSON validated; zero outputs and null `execution_count` throughout.
- `float()` coercion in the ML cell confirmed necessary: with the gold columns now genuinely `DECIMAL`, `F.avg(...)` returns `Decimal` and the existing `* 10` / `+ 50` arithmetic would raise `TypeError` on mixed `Decimal`/`float`.

**Not yet done:** a Run All on a live cluster. I have no workspace where I can execute this end to end, so outputs are cleared rather than committed from a partial run. The schema reasoning above is deterministic, but the cluster run is the real check and I would rather say so than imply it happened. Happy to take corrections from anyone who can run it.

**Test configuration**: Spark 3.5.0 · Delta 3.2.0-oci-1.0.0 · `spark.sql.sources.default=delta` · catalog impl `hive`

# Checklist:

- [ ] My changes generate no new warnings — not verifiable without a cluster run
- [ ] I have added tests that prove my fix is effective — this is a notebook sample; the assertions are the notebook's own printed counts

# Note for reviewers

The generator-vs-DDL mismatch in item 1 is not specific to the telecommunications variant — the same pattern appears in several of the other industry medallion demos. I have deliberately kept this PR to one notebook so the fix is reviewable; if the approach looks right I will open a follow-up for the rest rather than widening this one.


navid-heydari:
Closing in favour of #116, which carries this change as its first commit (`f00ac82`) along with the other three. Consolidated into a single PR at @ahmedawan-oracle's suggestion — the four commits are kept separate there so each concern is still reviewable on its own.

PR #111, opened 2026-09-15
fix(scd): make SCD2 change detection NULL-safe and parameterise the catalog

navid-heydari:
# Description

Four problems in `slowly_changing_dimension_template.ipynb`. The first is a correctness bug that loses history silently; the rest are what stopped the notebook from running as written.

**1. Change detection was not NULL-safe — history was silently dropped.**

The template rendered `target.{{ col }} != source.{{ col }}`. In Spark, `NULL != 'x'` evaluates to `NULL`, not `true`, and `NULL` is falsy in both `WHEN MATCHED AND` and `WHERE`. So a `NULL` → value transition (or the reverse) matched neither the expire step nor the insert step: the row was never expired and never re-inserted, and the change vanished with no error and no warning. For a dimension where any tracked column is nullable — which is most of them — this is a quiet data-loss bug, and the kind that is hardest to notice because the pipeline reports success.

Rendering `NOT (target.{{ col }} <=> source.{{ col }})` fixes it; `<=>` treats `NULL` as a comparable value. (`IS DISTINCT FROM` has identical semantics if you prefer it.)

Confirmed by emulating Spark's three-valued logic over the four staging rows:

| | `!=` (before) | `<=>` (after) |
|---|---|---|
| expired | `[1]` | `[1, 4]` |
| inserted | `[1, 3]` | `[1, 3, 4]` |

Customer 4 is the `NULL` → value case. Before the fix it appears in neither list.

**2. The target table was a hardcoded `lake.scd` that nothing in the notebook created.** Three cells failed at analysis time on any tenancy without a catalog literally named `lake`. There is now a configuration cell with `CATALOG` / `SCHEMA` / `TARGET_TABLE` / `SOURCE_VIEW`, it creates the catalog and schema if absent, and the `DROP TABLE` moved *after* the `CREATE SCHEMA` it depends on.

**3. `%pip install jinja2` removed.** There is already a `requirements.txt` next to this notebook listing jinja2; the prerequisites section now points at it.

**4. The template was not actually generic.** The insert step hardcoded `target.customer_id IS NULL`, so the "no current row" test only worked for this one dimension. `key_column` is now a parameter. The `print` calls that rendered the SQL were commented out — meaning they could not have produced the output that was committed — and are replaced by a `verbose` flag.

**Demo data now exercises all four paths.** Previously two of the four SCD2 branches were untested by the sample: nothing was unchanged and nothing was `NULL`. The target seeds Dana with a `NULL` email, and Bob is genuinely identical in staging:

| Customer | Path | Expected |
|---|---|---|
| Alice | tracked column changed | expire + insert |
| Bob | identical to target | no-op, stays on the original row |
| Charlie | no current row | insert only |
| Dana | `NULL` → value on email | expire + insert *(dropped before this fix)* |

A precondition that was implicit is now stated in the header: the MERGE expires one current row per key, so staging must hold at most one row per SCD key. Duplicates fail with a multiple-match error.

## Type of change

# How Has This Been Tested?

- **The NULL-safety bug and its fix were reproduced outside Spark**: the rendered SQL from both the old and new templates was evaluated against the four staging rows under SQL three-valued logic, producing the table above. This is the evidence for item 1 and it is reproducible without a cluster.
- Both templates rendered with jinja2 and the emitted SQL inspected for the parameterisation changes.
- Re-running the merge a second time is a no-op: after step 1 expires the changed rows, step 2 re-inserts them, and a third pass finds nothing to change.
- Notebook JSON validated; zero outputs and null `execution_count` throughout.

**Not yet done:** a Run All on a live cluster. Outputs are cleared rather than committed from a run I could not perform. The logic above was verified by emulation, not by Spark itself.

**Test configuration**: Spark 3.5.0 · Delta 3.2.0-oci-1.0.0 · jinja2 from the directory's `requirements.txt`

# Checklist:

- [ ] My changes generate no new warnings — not verifiable without a cluster run
- [ ] I have added tests that prove my fix is effective — this is a notebook sample; the expected result per customer is stated in-cell and printed by the final query


navid-heydari:
Closing in favour of #116, which carries this change as commit `54b85d1` along with the other three. Consolidated into a single PR at @ahmedawan-oracle's suggestion — the four commits are kept separate there so each concern is still reviewable on its own.

PR #114, opened 2026-09-15
Add data lineage sample: plan-derived lineage plus API conformance harness

### LINKED ISSUE #112: Add a data lineage sample, and a new "Data Engineering — Lineage" catalog section for it
Opening this per CONTRIBUTING's *Adding a New Example* step 3 — "If no existing section fits, discuss adding a new one in the issue before opening the pull request" — because the sample I want to contribute does not fit any existing section.

## What the sample is

A notebook that derives table- and column-level lineage from Spark's Catalyst analyzed plan and verifies it against a known DAG, plus a pytest harness that confirms AIDP's lineage API is released and enforces its request contract.

The notebook stands alone and needs no API access. The pytest module is a maintainer conformance harness: it signs real requests with an OCI profile and requires `AIDP_DATALAKE`, which has no default so an unconfigured run stops with a message rather than targeting someone else's instance.

## Why it needs a new section

CONTRIBUTING lists these homes under data engineering:

- `data-engineering/ingestion/` — connectors and loading patterns
- `data-engineering/transformation/` — pipeline architectures and table formats

Lineage is neither. It does not move or reshape data; it inspects what a pipeline *did*. Filing it under `transformation/` would put a verification tool in a directory of pipeline templates, and `ingestion/` is plainly wrong.

Proposal: `data-engineering/lineage/`, with a matching `### Data Engineering — Lineage` section in the Sample Catalog, placed between Ingestion and Transformation. If it lands, CONTRIBUTING's directory list should gain the same entry so the next contributor does not have to ask.

**Alternative if you would rather not add a section:** `data-engineering/transformation/lineage/` as a subdirectory, with the row going into *Other Transformation Patterns*. That needs no new section and no CONTRIBUTING change. I do not think it is the better home, but it is a smaller ask and I am happy to take it.

## One thing to know before reviewing

The sample documents an open gap rather than papering over it: on the tenancy I tested, `fetchLineage` is reachable and enforces its contract, but rejects every `anchorNode` form I could construct with `400 Invalid anchorNode`, so no graph comes back. The README leads with that, and the four affected tests are `xfail(strict=False)` so they flip to XPASS if the graph becomes populated rather than being deleted or skipped.

If that gap means this should not ship as a public sample until it is resolved, say so and I will close the PR — better to ask now than after review. The plan-derived notebook half does not depend on the API and is unaffected either way.

## Status

The PR is open so the code is there to look at, but **the directory question is the blocking one and belongs here.** Moving the sample to `transformation/lineage/` is a rename, so answering here costs nothing and saves a review round. Please do not merge the PR before this is settled.

navid-heydari:
# Description

Adds `data-engineering/lineage/`: a notebook that derives table- and column-level lineage from Spark's Catalyst analyzed plan and verifies it against a known DAG, plus a pytest harness that confirms AIDP's lineage operations are released and enforce their request contract.

Fixes #112

**Please read #112 before merging.** It asks the one question that belongs there rather than here: whether `data-engineering/lineage/` should be a new top-level section at all, or whether this should live at `data-engineering/transformation/lineage/` instead. CONTRIBUTING's *Adding a New Example* step 3 says to settle that in the issue first. Moving it is a rename, so answering costs nothing and saves a review round. This PR assumes the new section; say the word and I will move it.

## What is in it

| File | What it is |
|---|---|
| `Verify_Data_Lineage.ipynb` | Stands alone. Needs no API access. Derives lineage from the analyzed plan and checks it against a DAG it builds itself, with negative controls. |
| `test_aidp_lineage_api.py` | A **maintainer conformance harness**, not a self-contained demo. Signs real requests with your OCI profile. |
| `README.md` | Leads with the open gap below, then prerequisites, then scope. |
| `pytest.ini`, `requirements.txt` | Markers (`existence`, `population`) and pinned deps. |

## Two things stated up front rather than buried

**1. The API is released; the graph is not populated on the tenancy I tested.** `fetchLineage` is reachable and reaches its own parameter validation, but rejects every `anchorNode` form I could construct with `400 Invalid anchorNode` — the table key as `ListTables` returns it, the bare name, catalogGuid-qualified, storage-path, type-prefixed, and the DataLake OCID. Omitting the field instead returns `anchorNode must not be null`, a *different* error, so the server distinguishes missing from unresolvable: it is doing a real lookup and finding nothing.

Two explanations I could not separate: the graph genuinely is not populated for this DataLake (possibly lineage registers a process node only for Job/Workflow execution, not interactive notebook cells), or `anchorNode` wants an internal Data Catalog node id whose format is undocumented — the CLI reference lists the field with an empty description.

The four affected tests are `xfail(strict=False)` rather than deleted or skipped, so they flip to **XPASS** the moment the gap closes and the suite reports it. `test_B0` prints the full candidate matrix on every run.

**If this gap means the sample should not ship publicly yet, say so in #112 and I will close this.** I would rather ask than have it found in review.

**2. `test_aidp_lineage_api.py` requires your own tenancy.** `AIDP_DATALAKE` has deliberately **no default** — a default would point every reader's signed requests at someone else's resource and fail with an ambiguous 404. Unset, the suite stops with a message naming the variable. The README documents every variable with its default.

## Scope, stated in the README

The notebook's 19/19 is scoped to **writes whose top plan node is a `Project` or an `Aggregate`**, which covers aliases, aggregates, `COALESCE`/`CASE` across two sources, `SELECT *` expansion and self-joins. The README carries a table of the shapes that are *not* handled and how each one fails — window functions, `Project` over `Aggregate`, `LATERAL VIEW explode`, the second branch of a `UNION`, `Sort`/`Distinct`/`Filter`/`WithCTE` roots, CTEs, and subquery sources. Several of those return an empty column map *without warning*, which is exactly why the limit is documented rather than left to be discovered. It is a worked demonstration of the technique on a known DAG, not production tooling.

## Why the negative controls matter

An extractor that reported *every* table would satisfy "no missing edges" while being useless. So the notebook also asserts that a decoy table appears in no edge, that there is no direct `mart → raw_orders` edge (it is transitive only), and that no unresolved non-catalog leaves leak in. Same idea in the pytest suite: `A2` proves a fake sibling action really does 404, which is what gives `A1`'s 400 meaning, and `A6` proves the server actually parses `LineageDirection` — a stub that ignored the body and always complained about `anchorNode` would pass every other existence test and fail that one.

## Service naming

AIDP ships lineage under **`DataLineage`** (`DataLineageClient`, CLI group `data-lineage`). It shipped as `SemanticCatalog` in SDK v4.1.0 (2026-08-07) and was renamed by v4.2.1, so the README pins the name to the version. The operations are on `datalake.{region}` + `/20260430`, not the older `aidp.{region}` + `/20240831` — probing the old host yields a 404 that is easy to misread as "AIDP has no lineage API", and worse, on that surface the 404 for an absent route is byte-identical to the one for a denied permission. `A2` and `A7` encode both halves of that trap.

## Type of change

- [x] This change requires a documentation update — adds the Sample Catalog row and the repository structure entry; if the new section is accepted, CONTRIBUTING's directory list should gain `data-engineering/lineage/` too

# How Has This Been Tested?

- `test_aidp_lineage_api.py`: **13 passed / 4 xfailed** against a live AI Data Platform instance in us-ashburn-1. The 4 xfails are the population gap above. Reproduce with your own `AIDP_DATALAKE`: `pytest test_aidp_lineage_api.py -v` (add `-rX` to see why Part B is blocked).
- The notebook's 19 checks passed on Spark 3.5.0 / Delta 3.2.0-oci-1.0.0, including the three negative controls.
- `test_aidp_lineage_api.py` byte-compiles; notebook JSON validated.

**One caveat on the notebook, stated plainly:** it ships with **outputs cleared**. The run that produced the 19/19 happened before I removed an object-storage URI from one cell's output, which left the committed outputs no longer matching the committed code — one cell blank while a later cell claimed its results, and non-monotonic execution counts from re-runs. Rather than commit a set of outputs that misrepresents a run, I cleared them all. The 19/19 above is what I observed; it is not what this file currently displays, and it deserves a fresh clean run before or shortly after merge.

**Test configuration**: Spark 3.5.0 · Delta 3.2.0-oci-1.0.0 · `spark.sql.sources.default=delta` · catalog impl `hive` · us-ashburn-1 · `oracle-samples/aidataplatform-sdk` v4.2.1

# Checklist:

- [x] I have added tests that prove my feature works — `test_aidp_lineage_api.py`, and the notebook's 19 in-line checks with negative controls
- [x] New and existing unit tests pass locally with my changes — 13 passed / 4 xfailed, as above


navid-heydari:
Closing in favour of #116, which carries this sample as commit `7d35533` along with the other three. Consolidated into a single PR at @ahmedawan-oracle's suggestion — the four commits are kept separate there so each concern is still reviewable on its own.

The directory question in #112 is unchanged and still the blocking one for this part.

PR #115, opened 2026-09-15
Add zero-copy clone and table metadata sample

### LINKED ISSUE #113: Add a zero-copy clone and table metadata sample under Other Transformation Patterns
Filing this ahead of the PR, matching how the other new samples here have been paired with an issue.

## What the sample is

`data-engineering/transformation/delta-clone-metadata/` — a notebook covering two Delta capabilities on AIDP, each step asserting an explicit expected count:

1. **Zero-copy clone.** `SHALLOW CLONE` creates a table whose transaction log points at the *source's* parquet files, so it is created in constant time regardless of table size. Writing to either side diverges copy-on-write while the other stays unchanged. `DESCRIBE DETAIL` and `DESCRIBE HISTORY` confirm the zero-copy property and the `CLONE` commit.
2. **Structure without data**, via `CREATE TABLE ... LIKE` and `CREATE TABLE ... AS SELECT ... WHERE 1=0`.
3. **Table and column metadata in SQL** — `TBLPROPERTIES` at table scope, column `COMMENT` at column scope, and a registry table for metadata you can query and join.

No new section needed: the row goes into the existing *Other Transformation Patterns* table, and the directory sits under `data-engineering/transformation/`, which CONTRIBUTING already lists.

## Why it might be worth having

Most of the value is in the operational gotchas, which are the sort of thing you otherwise find out the hard way:

- `VACUUM` on a cloned source can delete files a shallow clone still references, and Delta does not track that dependency for you.
- `SELECT ... FROM (DESCRIBE DETAIL t)` is a Spark 3.5 *parser* error, not a missing feature — so `DESCRIBE` forms have to run top-level and be projected with the DataFrame API.
- `CREATE TABLE ... LIKE` accepts at most a two-part name in the Spark 3.5 grammar.
- `owner` is a reserved table property.
- `DEEP CLONE` availability differs between the open-source Delta 3.2 grammar and Oracle's `3.2.0-oci` build.

## Scope I deliberately left out

The notebook shows that metadata is *attached*, not *enforced* — nothing reads a tag and redacts a column at query time. A generic tag-to-mask bridge needs per-type mask shapes and careful identifier handling, so sketching it unsafely here seemed worse than leaving it out. There is already a `data-engineering/transformation/masking/` directory; that is the right home for it, as a separate sample, if it is wanted.

Opening the PR alongside this.

navid-heydari:
# Description

Adds `data-engineering/transformation/delta-clone-metadata/` — a notebook and README covering Delta `SHALLOW CLONE` as a constant-time zero-copy clone, copy-on-write divergence from the source, structure-only clones, and attaching and reading table and column metadata in SQL. Every step asserts an explicit expected count.

Fixes #113

No new catalog section: the row goes into the existing *Other Transformation Patterns* table, and the directory sits under `data-engineering/transformation/`, which CONTRIBUTING already lists.

## What it covers

| Topic | What the notebook shows |
|---|---|
| Zero-copy clone | `SHALLOW CLONE` creates a table whose log points at the **source's** parquet files, so creation is constant-time regardless of table size. `DESCRIBE DETAIL` and `DESCRIBE HISTORY` confirm the zero-copy property and the `CLONE` commit. Writing to the clone diverges copy-on-write; the source stays unchanged. |
| Structure without data | `CREATE TABLE ... LIKE` and `CREATE TABLE ... AS SELECT ... WHERE 1=0`. |
| Metadata in SQL | `TBLPROPERTIES` at table scope, column `COMMENT` at column scope, and a registry table for metadata you can query and join. |

## The gotchas are most of the value

These are documented in the README because each one costs an afternoon otherwise:

- **`VACUUM` can break a shallow clone.** The clone references the *source's* files; vacuuming the source may delete files the clone still needs, and Delta does not track that dependency. Avoid vacuuming a cloned source, or raise its retention window.
- **Some `DESCRIBE` forms are not subqueryable.** `SELECT ... FROM (DESCRIBE DETAIL t)` is a Spark 3.5 **parser** error, not a missing feature. So `DESCRIBE DETAIL` / `DESCRIBE HISTORY` / `DESCRIBE` run as top-level statements and are projected with the DataFrame API.
- **`CREATE TABLE ... LIKE` takes at most a two-part name** in the Spark 3.5 grammar, which is why the configuration cell sets the current catalog and schema with `USE` rather than fully qualifying every reference.
- **`owner` is a reserved table property.** The notebook uses `data_owner`.
- **`DEEP CLONE` availability differs by build** — the open-source Delta 3.2 grammar rejects it; Oracle's `3.2.0-oci` build may accept it.

## Two deliberate choices

**No try/except wrapper.** Statements execute directly, so anything unsupported on a given build fails at that cell rather than being swallowed and recorded as a "result". A harness that catches everything turns an unsupported feature into a passing row, which is the opposite of useful.

**`DEEP CLONE` is explained but not executed.** It follows from the point above: because availability is build-dependent, a cell that runs it unconditionally would halt Run All partway through on any build that rejects it. The markdown covers the difference from shallow clone and gives the statement to try against your own build.

## Scope I left out on purpose

The notebook is explicit that metadata is **attached, not enforced**: nothing reads a tag and redacts a column at query time. Writing a generic tag-to-mask bridge needs per-type mask shapes and careful identifier handling, and sketching that unsafely seemed worse than omitting it. `data-engineering/transformation/masking/` already exists and is the right home for it as a separate sample, if it is wanted.

## Type of change

- [x] This change requires a documentation update — adds the Sample Catalog row and the repository structure entry

# How Has This Been Tested?

The individual statements were exercised against Spark 3.5.0 / Delta 3.2.0-oci-1.0.0 while working out the grammar limits above — every gotcha in that list is something the build actually rejected, not something I read about. The notebook JSON is validated and it ships with outputs cleared.

**Not yet done:** a Run All of this file in its current, reorganised form. It was assembled from statements that ran individually, then restructured, and I have not re-executed it top to bottom since. Each step prints an expected count next to the actual one, so a reviewer running it will see immediately if anything disagrees.

**One thing I could not verify:** `CREATE TABLE customers_empty LIKE customers` does not pass `USING delta`. Spark's `CreateTableLikeCommand` should inherit the source's provider, so it ought to produce a Delta table, but I did not confirm the resulting format on this build. If it comes back as parquet, the cell needs an explicit `USING delta` — worth a glance from anyone with a cluster in front of them.

**Test configuration**: Spark 3.5.0 · Delta 3.2.0-oci-1.0.0 · `spark.sql.sources.default=delta` · catalog impl `hive`

# Checklist:

- [ ] My changes generate no new warnings — not verifiable without the full run noted above
- [x] I have added tests that prove my feature works — each step asserts an explicit expected count inline


navid-heydari:
Closing in favour of #116, which carries this sample as commit `f2786bc` along with the other three. Consolidated into a single PR at @ahmedawan-oracle's suggestion — the four commits are kept separate there so each concern is still reviewable on its own.

PR #116, opened 2026-09-15
Fix medallion and SCD notebooks, add data lineage and zero-copy clone samples

### LINKED ISSUE #113: Add a zero-copy clone and table metadata sample under Other Transformation Patterns

navid-heydari:
# Description

Repairs two existing notebooks and adds two new samples, plus test suites covering the two repairs. Supersedes #107 (closed), and addresses @ahmedawan-oracle's reviews of both #107 and this PR.

Fixes #113. Relates to #112 (whether `data-engineering/lineage/` should be a new top-level section — see [Open question](#open-question-112) below).

**Relative to `main`:** 5 files modified, 11 added, 0 removed.

| Modified | |
|---|---|
| `transformation/medallion-lake/telecommunications_medallion_architecture_demo.ipynb` | bronze/silver/gold now conform to their DDL (25 → 26 cells) |
| `transformation/scd/slowly_changing_dimension_template.ipynb` | NULL-safe change detection, parameterised target (10 → 12 cells) |
| `transformation/scd/requirements.txt` | local-test deps, marked not-required-on-AIDP |
| `README.md` | Sample Catalog rows + repository-structure tree |
| `CONTRIBUTING.md` | `data-engineering/lineage/` added to the directory list |

| Added | |
|---|---|
| `data-engineering/lineage/` | `README.md`, `Verify_Data_Lineage.ipynb`, `test_aidp_lineage_api.py`, `conftest.py`, `requirements.txt` |
| `transformation/delta-clone-metadata/` | `README.md`, `zero_copy_clone_and_table_metadata.ipynb` |
| `transformation/scd/` | `test_scd2_template.py`, `conftest.py` |
| `transformation/medallion-lake/` | `test_medallion_ddl_conformance.py`, `requirements.txt` |

Commits, one concern each:

| Commit | Change |
|---|---|
| `f00ac82` | Fix type mismatch and gold schema drift in the telecom medallion demo |
| `54b85d1` | Make SCD2 change detection NULL-safe and parameterise the catalog |
| `7d35533` | Add `data-engineering/lineage/` |
| `f2786bc` | Add `transformation/delta-clone-metadata/` |
| `7451336` | Address the four majors from the review of `f2786bc` |
| `6e3828c` | Address the minors and notes from the review of `f2786bc` |
| `f99e7ca` | Fix `-rX` and remaining investigation voice in the harness docstring |
| `10591dd` | Add test suites for the SCD template and medallion DDL conformance |

The last four are review responses and are natural squash candidates.

---

## 1. Medallion — the notebook could not complete a Run All

**The bronze write failed on a type mismatch.** `createDataFrame()` infers `double` for a Python float and `bigint` for an int, while the bronze DDL declares `DECIMAL(10,3)`, `DECIMAL(8,2)` and `INT`. A Delta `overwrite` through `saveAsTable` validates the DataFrame's schema against the table's and raises `DELTA_FAILED_TO_MERGE_FIELDS`, because decimal and double are not implicitly reconciled. The generating cell now casts each column to its declared type.

**The silver write failed for a different reason** — it carried an *extra* column (`ingestion_timestamp`, inherited from bronze) that the silver table does not declare. Silver now ends with an explicit `.select(...)` of exactly the DDL columns.

**The RCA markdown is rewritten** because the three mismatch cases do not behave alike:

| Case | Result |
|---|---|
| Type mismatch | **fails** — `DELTA_FAILED_TO_MERGE_FIELDS` |
| Column in the table, absent from the DataFrame | **succeeds**, written as `NULL` |
| Column in the DataFrame, absent from the table | **fails** |

**Each gold aggregate is `CAST` to its declared type**, `created_at` is projected, and columns are in DDL order. `SUM`/`AVG`/`COUNT` widen their results (`decimal(21,3)`, `double`, `bigint`), so without the casts the writes do not match the declared schema.

**`total_call_minutes` widened to `DECIMAL(12,2)`** on `network_infrastructure` and `service_performance` — the two tables that SUM it across a service type, where it reaches ≈1.24M against a `DECIMAL(8,2)` ceiling of 999,999.99. Under the default `spark.sql.ansi.enabled=false` that overflow is a **silent NULL**, not an error. `subscriber_analytics` stays at `DECIMAL(8,2)` deliberately: it is a per-subscriber total and cannot overflow. Headroom was computed from the generator's parameters for every summed column; the tightest of the rest has ≈7.6× margin at 10,000 subscribers.

**Tables are dropped before creation** in all three layers. A workspace that ran an earlier version holds them with the schema Spark inferred from the uncast frames, so `CREATE TABLE IF NOT EXISTS` would leave them alone and the corrected writes would fail the schema check on exactly the workspaces the fix is for.

**Two further real fixes, previously undocumented here:**

- **`processed_timestamp` is now populated.** `main` declares the column in the silver DDL and never writes it, so the audit field was always `NULL`.
- **`float()` wraps in the ML cell.** The gold columns are `DECIMAL`, so `.collect()` returns `Decimal`, and `Decimal * float` raises `TypeError`. This is a consequence of making gold honour its `DECIMAL` DDL, so it belongs with that fix.

**Also:** `primary_service_type` now has a `service_type` tie-breaker (`ROW_NUMBER ... ORDER BY COUNT(*) DESC` alone picks an arbitrary winner, which affects ≈13% of subscribers at 10k), and cell 10's `from pyspark.sql.functions import *` is replaced with explicit imports — the wildcard bound pyspark's `round`/`max`/`min` over the builtins the generator cell calls on plain numbers.

## 2. SCD — change detection dropped history silently

The template rendered `target.{{ col }} != source.{{ col }}`. In Spark, `NULL != 'x'` is `NULL`, which is falsy in both `WHEN MATCHED AND` and `WHERE`. A `NULL` → value transition matched neither the expire step nor the insert step: the row was never expired, never re-inserted, and the change vanished with no error. Rendering `NOT (target.{{ col }} <=> source.{{ col }})` fixes it.

| | `!=` (before) | `<=>` (after) |
|---|---|---|
| expired | `[1]` | `[1, 4]` |
| inserted | `[1, 3]` | `[1, 3, 4]` |

Customer 4 is the `NULL` → value case; before the fix it appears in neither list. **This is now executed, not argued** — see [Tests](#tests).

Also in this commit:

- **The target is parameterised.** `main` hardcodes `hive.default.dim_customer`. There is now a configuration cell with `CATALOG`/`SCHEMA`/`TARGET_TABLE`/`SOURCE_VIEW` that creates both, with the `DROP` after the `CREATE SCHEMA` it depends on.
- **`key_column` is now a parameter.** The insert step hardcoded `target.customer_id IS NULL`, so the "no current row" test only worked for this one dimension.
- **The `print` calls that rendered the SQL shipped commented out** on `main`, so the committed code could not have produced the committed output. Replaced with a `verbose` flag.
- **Demo data now exercises all four paths.** Two were previously untested: nothing was unchanged and nothing was `NULL`.

| Customer | Path | Expected |
|---|---|---|
| Alice | tracked column changed | expire + insert |
| Bob | identical to target | no-op, stays on the original row |
| Charlie | no current row | insert only |
| Dana | `NULL` → value on email | expire + insert *(dropped before this fix)* |

- **The staging uniqueness precondition is enforced, not just documented.** It fails two ways: for a key that already has a current row, step 1 aborts with a multiple-match error; for a key with *no* current row, nothing fails and the dimension silently gains two current rows. `run_scd2_merge` now refuses to run.

## 3. New sample — `data-engineering/lineage/`

A notebook that derives table- and column-level lineage from Spark's Catalyst analyzed plan and verifies it against a known DAG, plus a pytest harness confirming AIDP's lineage operations are released and enforce their request contract. The notebook stands alone and needs no API access.

**The API is released; no graph came back on the tenancy tested.** `fetchLineage` is reachable and reaches its own parameter validation, but rejected every `anchorNode` form I could construct with `400 Invalid anchorNode`. Omitting the field instead returns `anchorNode must not be null` — a *different* error, so the server distinguishes missing from unresolvable and is doing a real lookup. Either the graph is not populated for that DataLake, or `anchorNode` wants an internal Data Catalog node id whose format is undocumented; I could not separate the two. Observed **2026-08-16, one DataLake in us-ashburn-1, SDK v4.2.1** — dated throughout, because this is one observation and not a statement about the service. The four affected tests are `xfail(strict=False)` so they flip to **XPASS** if your tenancy differs.

**Part A does not depend on that gap persisting.** Part A probes with a deliberately unresolvable sentinel and Part B passes the real anchor explicitly, so the suite reports 13 passed / 4 xfailed today and 13 passed / 4 **xpassed** the day the graph is populated. `AIDP_ANCHOR_TABLE` feeds Part B only.

**`AIDP_DATALAKE` has no default.** Unset, the suite stops with a message naming the variable — a collection error at import, deliberately, because a session fixture calling `pytest.fail` would be absorbed by Part B's `xfail` markers and look like the known gap.

**The accuracy claim is scoped to what is proven.** The 19/19 covers writes whose top plan node is a `Project` or an `Aggregate` **and where every attribute that node references is emitted directly by a leaf relation**. That second condition is the real limit: `colmap` is built only from `collectLeaves()`, so an alias minted by an intermediate `Project` resolves to nothing regardless of the top node. The README carries a table of the unhandled shapes. It is a worked demonstration of the technique, not production tooling.

**Service naming:** `DataLineage` (`DataLineageClient`, CLI group `data-lineage`). It shipped as `SemanticCatalog` in SDK v4.1.0 and was renamed in **v4.1.1 (2026-08-31)**, flagged breaking in the SDK changelog.

<a name="open-question-112"></a>
**Open question (#112):** whether `data-engineering/lineage/` should be a new top-level section, or belongs at `data-engineering/transformation/lineage/`. CONTRIBUTING's *Adding a New Example* step 3 says to settle that before opening a PR. Moving it is a rename — please answer there and I will do it. **This gates the lineage commit only**; if it stalls, I would rather peel `f00ac82` + `54b85d1` (plus `10591dd`) into a fixes-only PR than hold verified bug fixes behind a naming decision.

## 4. New sample — `transformation/delta-clone-metadata/`

Replaces the `capability-probes/` notebook from #107. Covers Delta `SHALLOW CLONE` as a constant-time zero-copy clone, copy-on-write divergence, structure-only clones, and attaching and reading table and column metadata in SQL. Each step **asserts** its expected value, so a Run All that disagrees fails at the cell that disagreed. No new catalog section: the row goes into the existing *Other Transformation Patterns* table.

**Zero-copy is proven properly.** `DESCRIBE DETAIL` alone does not demonstrate it — `format`/`numFiles`/`sizeInBytes` read the same for a deep copy, and `DESCRIBE HISTORY` reports `operation = CLONE` for either. The notebook asserts the `CLONE` commit's `operationMetrics.numCopiedFiles == 0` and that the clone's `inputFiles()` are exactly the source's, then shows them diverging after a write.

**Every reference is a three-part name, and `USE` is not used.** `CREATE TABLE ... LIKE` is the v1 `CreateTableLikeCommand`; Spark 3.5 and Delta 3.2 both resolve its source through the v1 `SessionCatalog`, and `CatalogManager.setCurrentNamespace` only syncs the v1 current database when the current catalog is `spark_catalog`. Against a plugin catalog the `USE` + `LIKE` pair raises `TABLE_OR_VIEW_NOT_FOUND`, or silently binds to a same-named table in `default`. The `LIKE` example is replaced by a CTAS with an explicit `USING delta`.

**Other gotchas documented:**

- `SELECT ... FROM (DESCRIBE DETAIL t)` is a Spark 3.5 **parser** error, not a missing feature. The `DESCRIBE` family runs top-level and is projected with the DataFrame API.
- `owner` is a reserved table property; the notebook uses `data_owner`.
- `DEEP CLONE` is explained but **not executed** — availability differs between the OSS Delta 3.2 grammar and Oracle's `3.2.0-oci` build.
- **`VACUUM` on a cloned source can delete files a shallow clone still references.** The repair is to re-clone, but a bare `CREATE OR REPLACE TABLE ... SHALLOW CLONE` over a clone holding its own data raises `DELTA_UNSUPPORTED_NON_EMPTY_CLONE`, so `DELETE FROM` it first.

**The `run()` harness from #107 is gone.** It swallowed `ParseException` messages, printed Py4J errors without detail, and recorded unexecuted statements as SUPPORTED.

**Left out on purpose:** the generic tag-to-mask bridge. Metadata here is *attached, not enforced*; `transformation/masking/` already exists and is the right home for it.

<a name="tests"></a>
## 5. Tests for the two notebook repairs

Added in `10591dd`, following the convention already in `transformation/merge/` (pytest, a local `conftest.py`, and a `requirements.txt` marking which deps are local-test-only).

### `transformation/scd/test_scd2_template.py` — 13 tests

The templates and `run_scd2_merge` are lifted out of the notebook's own cell rather than duplicated, so the tests fail if the notebook drifts.

| Tier | Needs | Proves |
|---|---|---|
| Rendering | jinja2 | every tracked column compared with `<=>`, no bare `!=`, `key_column` genuinely parameterised |
| Predicate semantics | local Spark, no Delta | the **rendered predicate text** evaluated in Spark's real three-valued logic across 5 cases, plus a test showing the `!=` form returns `NULL` and is therefore falsy in `WHERE` |
| Guard | none | duplicate staging keys raise **before** either statement executes, and the message names the keys |
| End-to-end (`-m delta`) | delta-spark | the full expire-then-insert pair: expired `[1, 4]`, exactly one current row per key, Bob unmoved, second run a no-op |

The end-to-end tier is the first *execution* of this notebook's logic — the table in section 2 was previously established by emulation on both sides of the review.

Delta's extensions are configured on the single shared session up front. Spark permits one active session per JVM, so a later `getOrCreate()` returns it; building a plain session first made the Delta test fail with `Py4JJavaError`, but **only when both suites ran in the same process**. It passed in isolation, which is the failure mode this review has been about.

### `transformation/medallion-lake/test_medallion_ddl_conformance.py` — 8 tests + 12 cases

Pure notebook-JSON parsing: no Spark, no cluster, no network. Extracts the narrow-typed columns from each bronze `CREATE TABLE` and checks the generating cell casts to them.

Parameterised over all 12 demos. The six that still carry the defect — `energy`, `financial_services`, `manufacturing`, `real_estate`, `retail`, `transportation` — are `xfail(strict=False)`, so each flips to **XPASS** as it is fixed and the suite reports the gap closing rather than needing the list hand-maintained. Verified by simulating a cast fix in `retail`.

The `telecommunications` assertions enforce the reference fix and must not xfail: the per-table `DECIMAL` widths, no write passing `overwriteSchema`, bronze and silver dropped before creation, no wildcard function import, and the `primary_service_type` tie-breaker.

## Hygiene and process

- **Tenancy identifiers scrubbed**, and the notebooks no longer *produce* them: the lineage notebook prints only the path portion of a `DESCRIBE DETAIL` location, the clone notebook does not project `location`, and both READMEs ask contributors to clear outputs before committing. Built fresh on `main` rather than carrying `03c4cba`, so no OCID is in this PR's history.
- **Stale run records stripped.** The medallion notebook carried `command_metadata`/`execution`/`result_type` on every code cell from a 2026-07-31 run — three with `result_type: "error"` on the very cells this PR fixes — plus a `Last_Active_Cell_Index` of 27 on a 26-cell notebook. `main` carries none of it. The workbench `type` metadata that `main` *does* use is kept.
- **`TAKEAWAYS.md` is gone**, along with the field-sales voice and the "one question for Oracle" framing.
- **`pytest.ini` became `conftest.py`.** An ini file is only read as the rootdir config, so running `pytest data-engineering/lineage/` from the repository root silently ignored it and every marker came back an unknown mark.
- Every commit is signed off; the root README's Sample Catalog and repository-structure tree are both updated.

## Type of change

# How Has This Been Tested?

**Executed locally (Spark 3.5.2 / Delta 3.2.0, reproducible without a cluster):**

- `test_scd2_template.py` — **13 passed**. Both suites together from the repo root: **21 passed / 4 skipped / 6 xfailed**. With `-m "not delta"`: 12 passed. All four invocation orderings verified, because one of them originally failed.
- `test_medallion_ddl_conformance.py` — 8 passed, 4 skipped (demos with no narrow bronze types), 6 xfailed (the tracked demos).
- **The Part A / Part B anchor coupling was reproduced before it was fixed.** Against an in-process mock implementing only the documented behaviours, the pre-fix harness gives **7 failed / 6 passed / 4 xpassed** once the graph is populated (`A1`, `A3`, and all five `A5` parametrisations). The fixed harness gives 13 passed / 4 xfailed today and 13 passed / 4 xpassed populated.
- `conftest.py` verified to register its markers when pytest runs from the repository root, where the old `pytest.ini` produced unknown-mark warnings. The lineage suite now emits **zero** warnings.
- Every medallion write's projection reconciled against its table's DDL — names, order, and declared type vs. the type the expression produces. `DECIMAL` headroom computed from the generator's parameters for every summed column, which is how the `DECIMAL(8,2)` overflow surfaced.
- All four notebooks validate as nbformat JSON and every code cell byte-compiles.

**Verified on a cluster:**

- `test_aidp_lineage_api.py`: **13 passed / 4 xfailed** against a live AI Data Platform instance in us-ashburn-1 (2026-08-16). Reproduce with your own `AIDP_DATALAKE`: `pytest test_aidp_lineage_api.py -v` (`-rx` for xfail reasons; `-rP -k B0` for the anchor-candidate matrix).
- The lineage notebook's 19 checks passed on Spark 3.5.0 / Delta 3.2.0-oci-1.0.0 (2026-08-16), including the three negative controls.
- Every gotcha in section 4 is something the build actually rejected while the statements were being worked out individually.

**Not yet done, and why all four notebooks ship with outputs cleared:**

None of the four has had a **Run All on an AIDP cluster** in its current form. I have no workspace to execute them in, so committing outputs would mean committing a partial run. The lineage notebook's 19/19 is real, but that run predates removing the object-storage URI from one cell's output, which left the committed outputs no longer matching the committed code. Rather than ship outputs that misrepresent a run, I cleared all four. Every step prints or asserts an expected value next to the actual one.

The new suites narrow this gap but do not close it: the SCD repair is now executed end-to-end against Delta, while the medallion repair is verified **statically** against its DDL rather than by running the notebook.

**Test configuration**: Spark 3.5.0 · Delta 3.2.0-oci-1.0.0 · `spark.sql.sources.default=delta` · catalog impl `hive` · us-ashburn-1 · SDK v4.2.1. Local suites: Spark 3.5.2 · Delta 3.2.0 · Python 3.11.

# Checklist:

- [ ] My changes generate no new warnings — the Python suites emit zero warnings (the `conftest.py` change removed 13 unknown-mark warnings); notebook-execution warnings are not verifiable without the cluster Run All noted above
- [x] I have added tests that prove my fix is effective — lineage (`test_aidp_lineage_api.py` + 19 in-notebook checks), clone (an `assert` per step), SCD (`test_scd2_template.py`, including an executed Delta end-to-end), medallion (`test_medallion_ddl_conformance.py`, **static** conformance rather than execution)
- [x] New and existing unit tests pass locally with my changes — 21 passed / 4 skipped / 6 xfailed for the two new suites, 13 passed / 4 xfailed for the lineage suite. Note `.github/` contains only `pull_request_template.md`, so no CI runs tests on PRs

# Note for reviewers

The generator-vs-DDL mismatch in section 1 is not specific to the telecommunications variant. I confirmed it independently: of the 12 medallion demos, **6 others** declare narrow bronze types and write an uncast `createDataFrame` frame — `energy`, `financial_services`, `manufacturing`, `real_estate`, `retail`, `transportation`. (`insurance` already casts; `education`, `healthcare`, `hospitality`, `media` declare no narrow bronze types.) This PR touches one notebook deliberately, so the fix stays reviewable — and `test_medallion_ddl_conformance.py` now encodes the other six as `xfail`, so the follow-up work reports its own progress. Tracking issue to follow.


ahmedawan-oracle:
## Deep review of `f2786bc` — the #107 items are verified fixed by execution across all four notebooks; four majors remain: the clone sample's `USE` + `CREATE TABLE … LIKE` pattern cannot work on upstream Spark with a plugin catalog and has never been run, the lineage harness's Part A asserts 400 on the very anchor Part B expects to resolve, the location print that leaked the namespace in #107 is still there (now in two notebooks), and the narrowed lineage scope names the wrong condition

@navid-heydari — this is a serious, careful answer to the #107 review and most of it checks out under execution: the medallion notebook now conforms to its DDL in all three layers, the SCD template is NULL-safe and re-runnable, the identifiers are gone from files and history, every commit is signed off, and the docs are in a public-sample register. Method as before: everything below was established by execution where this Windows host allows it (Spark 3.5.9 + Delta 3.2.0 parser and in-memory semantics, Delta's own `SchemaMergingUtils.mergeSchemas` for every write, a non-session v2 catalog for the clone sample's `USE`/`LIKE` path, the pytest module in a network-isolated sandbox with an in-process mock server, the extractor on 37 query shapes, the templates rendered and emulated), and from Spark/Delta 3.2 sources and the public SDK for the rest; catalog writes cannot run here and no live AIDP call was made, so write-dependent behaviour is labelled. Every finding survived independent adversarial re-verification (44/44). Also: #107 is still open with your question ("split into four PRs or fix in place?") unanswered — I have replied there; see the bottom.

### Verified fixed since #107 ✔

- **Medallion** (executed against the DDL with Delta's merge check): bronze cell 5 casts + explicit select turn `createDataFrame`'s alphabetical double/bigint frame into exactly the bronze DDL (main's uncast frame throws `DELTA_FAILED_TO_MERGE_FIELDS` on `data_volume`); silver cell 10 selects exactly the 12 DDL columns and now populates `processed_timestamp` (without the select the merge yields 13 fields, which Delta refuses — the #107 silver failure); all three gold queries produce every column at the declared type, in DDL order, with `created_at`, and `overwriteSchema` is gone. The rewritten RCA (cell 6) is correct on all three rows. The `DECIMAL(12,2)` widening is justified: Voice `total_call_minutes` per service type ≈ 1,237,500 > 999,999.99, and under the default `spark.sql.ansi.enabled=false` that overflow is a silent NULL (executed: `CAST(1237500.00 AS DECIMAL(8,2))` → NULL). No other column overflows at 10,000 subscribers (7.6× headroom on the tightest). The ML cells type-check against the DECIMAL gold schema, and the `float()` wraps in cell 24 are a necessary third fix (`Decimal * float` raises) — worth a line in the body.
- **SCD** (emulated with the rendered predicate text): `NOT (a <=> b)` gives expired [1,4] / inserted [1,3,4] vs `!=`'s [1] / [1,3] — exactly your table; value→NULL also expires+inserts; NULL→NULL is a no-op; a second run is a no-op (0/0, 6 rows); Bob stays on his original row; `key_column`, `verbose`, the config cell, the `DROP` after `CREATE SCHEMA`, the removed `%pip`, `requirements.txt` beside the notebook, and the `IS DISTINCT FROM` remark all check out; `datetime.datetime` is accepted for `DateType`.
- **Lineage**: the OCID default is gone (unset → `RuntimeError` naming the variable); outputs cleared and nbformat-valid; README 109–111 is exactly what happens without a profile (13 fixture ERRORs, 4 XFAIL, zero requests sent); 17 items = 13 + 4; the six anchor forms, env-var table, requirements and OpenLineage guidance all match; every row of the new Scope table (195–198) matches executed behaviour; the "whole SDK moved" framing is gone; the demo's five column expectations resolve.
- **Clone**: all 22 statements parse on Spark 3.5.9 + Delta 3.2; the five #107 malformed statements and the `run()` harness are gone; `DEEP CLONE` really is a `PARSE_SYNTAX_ERROR` in the OSS 3.2 grammar; `SELECT … FROM (DESCRIBE DETAIL t)` really is a parser error; `owner`/`location`/`provider` are reserved and `classification`/`data_owner`/`pii` are not; the projected `DESCRIBE DETAIL`/`HISTORY` columns exist; the VACUUM warning matches the Delta docs; your open question is answered — with a Delta source `CREATE TABLE … LIKE` is intercepted by Delta's `EligibleCreateTableLikeCommand` and yields provider `delta` regardless of `spark.sql.sources.default` (Delta ≥ 2.3), so no `USING delta` is needed for that (but see major 1).
- **Hygiene/process**: no OCID, namespace, bucket or tenancy name in any file, output or cell metadata; `03c4cba` is not an ancestor and `git log -S ocid1.` over the range hits only the two placeholder lines; Signed-off-by on all four commits; template filled with three boxes honestly unticked; root README tree/section/rows alphabetical and per #112's proposal; TAKEAWAYS, capability-probes and the field-sales language are gone; #112/#113 exist.

### Majors

1. **`USE default.clone_metadata_demo` cannot make `CREATE TABLE customers_empty LIKE customers` work on upstream Spark with a plugin catalog** (`zero_copy_clone_and_table_metadata.ipynb` cells 1 and 9; README:32; cell 8). `CREATE TABLE … LIKE` is the v1 `CreateTableLikeCommand`; both Spark 3.5 and Delta 3.2's `EligibleCreateTableLikeCommand` resolve the source through the v1 `SessionCatalog`, and `CatalogManager.setCurrentNamespace` only syncs the v1 current database when the current catalog is `spark_catalog`. Executed on Spark 3.5.9 + Delta 3.2.0 with a non-session v2 catalog standing in for AIDP's plugin catalogs: `CREATE SCHEMA` ✓, `USE testcat.clone_metadata_demo` ✓, `CREATE TABLE customers` ✓ (`count()` = 2), then `CREATE TABLE customers_empty LIKE customers` → `[TABLE_OR_VIEW_NOT_FOUND] default.customers`, while the CTAS in the same cell succeeds; if a `customers` temp view or v1 table exists, LIKE silently binds to that instead. On the stock session catalog cell 1 fails first (`USE default.clone_metadata_demo` → `SCHEMA_NOT_FOUND`), so the notebook is AIDP-only — and the PR body says it has never had a Run All and the LIKE result was never inspected; no sibling notebook uses `USE` or `LIKE` (0/155), and Oracle's own SQL grammar page shows every table statement fully qualified. Consequence: Run All most likely halts at cell 9 and the cleanup cell never runs, while README:32 and cell 8 teach `USE` as the correct workaround. Fix: drop the LIKE example (keep the CTAS with an explicit `USING delta`, or show `CREATE OR REPLACE TABLE c.s.customers_empty SHALLOW CLONE c.s.customers VERSION AS OF 0`, which parses 3-part), remove the `USE`, fully qualify every statement (all of them parse 3-part — executed), and delete the README:32 bullet and cell 8's sentence. If you can paste `DESCRIBE DETAIL default.clone_metadata_demo.customers_empty` from an AIDP Run All showing `format = delta` in that schema, this drops to "document the AIDP-only dependency".
2. **Part A of the lineage harness asserts 400 for the same anchor Part B expects to return 200** (`test_aidp_lineage_api.py`: `fetch_lineage()` defaults `anchorNode` to `ANCHOR_TABLE` at 96–105; A1 139–144, A3 165–171, A5 211–215 assert 400 / `Invalid anchorNode` on that body; B1–B4 assert 200 on the identical body). Executed against an in-process mock that implements only the behaviours the module and README document: unpopulated → your 13 passed / 4 xfailed; the moment the default anchor resolves → **7 failed / 6 passed / 4 xpassed** (A1, A3, A5×5 fail with "expected 400 … got 200"). That contradicts the docstring ("Part A … must be green", Part B "flip to XPASS and the suite tells you the gap closed"), README:130–136, and README:111 then tells the reader to distrust the XPASS. `AIDP_ANCHOR_TABLE` (README:107) is a today-path to the same breakage on any tenancy where lineage is harvested. A4 already uses the right pattern. Fix (12 lines, verified: 13 passed / 4 xfailed today, 13 passed / 4 xpassed populated): default `fetch_lineage()`'s `anchorNode` to an unresolvable sentinel, use the sentinel in A3's `exportLineage` body, pass `anchorNode=ANCHOR_TABLE` explicitly in B1–B3.
3. **The location print that leaked the namespace and managed bucket in #107 is unchanged, and the new notebook adds a second one.** `Verify_Data_Lineage.ipynb` cell 18 is byte-identical to #107 and still does `print("     loc: %s" % loc)` — in #107 that exact cell's committed output was the sole source of the three `oci://<bucket>@<namespace>/…` lines; `zero_copy_clone_and_table_metadata.ipynb` cell 5 projects `location` from `DESCRIBE DETAIL` with `show(truncate=False)`. Outputs are cleared now (verified: zero `oci://` anywhere), but the body says a Run All is pending, the lineage commit message says outputs are cleared "pending a clean end-to-end run", and the repo has no nbstripout/CONTRIBUTING guard — so the next committed run republishes the runner's tenancy identifiers and the "scrubbed" claim becomes false. Fix: cell 18 → `print("     loc: .../%s" % loc.split("/", 3)[-1])` (verified to keep the `startswith` check and drop the authority) or print only the file count; clone cell 5 → drop `location` from the select; one README sentence asking contributors to clear outputs before committing.
4. **The narrowed lineage claim names the wrong condition.** README:185–188 and cell 23 say the extractor is accurate for "writes whose top plan node is a Project or an Aggregate … aliases, aggregates". Executed with the verbatim cell-4 code: `SELECT * FROM (SELECT cust_id, amount*2 AS amt FROM raw_orders) s` (top = Project) → `amt` ← (none); `SELECT cust_id, SUM(amt) … FROM (subquery) GROUP BY` (top = Aggregate) → `total` ← (none); `df.withColumn('x', …).withColumn('y', x+1)` → both empty; a JOIN to a derived subquery likewise — the top node is Project/Aggregate and the alias is plain, but the referenced attribute is minted by an intermediate Project and is absent from the leaf map, so cell 14 prints it as `(literal)`. The Scope table lists Project-over-Aggregate but not Project-over-Project / Aggregate-over-Project, the most common SQL shape after a flat SELECT. The true condition is "every attribute the top node references is emitted directly by a leaf relation". Fix: reword 185–188 to that, add a Scope row ("alias computed in a FROM-subquery or an earlier `.withColumn`/`.select` → that column's lineage is empty, shown as (literal)"), mirror in cell 23 and the body — or add a ~15-line transitive alias resolver to cell 4 and a WARN when `refs.size() > 0` and nothing resolved.

### Minors

- **Medallion — stale run record.** All 19 code cells still carry AIDP `command_metadata`/`execution`/`result_type` from the 2026-07-31 run of the 29-cell #107 tree, with `result_type: "error"` on cells 5, 11 and 24 — the bronze write, the silver write and the model cell this PR fixes — and notebook-level `Last_Active_Cell_Index: 27` on a 26-cell notebook; main has none of this (115 of the diff's 298 added lines are this noise). Strip cell metadata with the outputs (`nbstripout --extra-keys …` or a 3-line nbformat pass), keep kernelspec/language_info as main does.
- **Medallion — bronze/silver keep `CREATE TABLE IF NOT EXISTS` without `DROP`, though your gold rationale applies to them.** Main's committed outputs show a build where the uncast bronze/silver frames were written "successfully" over the DECIMAL DDL (and gold printed `revenue_potential` as a double), i.e. workspaces exist whose bronze/silver hold the inferred schema; executed `mergeSchemas(legacy double/long table, PR cast frame)` → `DELTA_FAILED_TO_MERGE_FIELDS` on `call_duration`. On those workspaces cell 5 fails with the very error cell 6 says the casts prevent. Add the same `DROP TABLE IF EXISTS` before cells 3 and 9 (or state in cell 6 that a fresh `telecom` catalog is assumed) and reword the cell-14 comment ("Dropped first, not CREATE TABLE IF NOT EXISTS" is followed by exactly that statement).
- **Medallion — `from pyspark.sql.functions import *` in cell 10 shadows `round`/`max`/`min`** used with Python numbers in cell 4; a forward Run All is fine, but re-running the generator after silver raises `PySparkTypeError … got float` (executed). Cell 5 already uses explicit imports; 74 of 77 data-engineering notebooks do.
- **SCD — the header's duplicate-key sentence is right for existing keys and wrong for new ones.** For a key that already has a current row, two staging rows make the expire MERGE fail (Delta counts every ON-matched source row, whether or not they differ — verified in `ClassicMergeExecutor.findTouchedFiles`); for a key with no current row nothing fails and the rendered INSERT writes two current rows (executed: staging [3,3] → two current rows, no error). Reword to say both halves; optionally guard in `run_scd2_merge` (`SELECT key … GROUP BY key HAVING count(*) > 1` must be empty).
- **Clone — "each step asserts an explicit expected count"** (body §4, checklist): no cell contains an `assert`; cells 4/6/9 print "(expect N)" beside the value, cells 5/11/12/14 only `.show()`. A Run All whose counts disagree finishes green. Either say "prints" and untick, or add one-line asserts plus expectations for cell 5 (`format == 'delta'`, history v0 `operation == 'CLONE'`) and cell 14 (2 rows).
- **Clone — `DESCRIBE DETAIL` of the clone alone does not demonstrate zero-copy**: `format/numFiles/sizeInBytes/location` would read the same for a deep copy, and the projected `DESCRIBE HISTORY` columns show `CLONE` for either. Delta records the proof in `operationMetrics` (`numCopiedFiles = 0`, `sourceNumOfFiles = N`) — add it to the cell-5 select, or compare `inputFiles()` of clone and source before the insert.
- **Clone — "Environment as tested" (README:44–46) and "tested on …" (cell 0)** for a notebook the body says has never been run end-to-end; in this repo that heading means an end-to-end run (`adw-ext-table-on-uniform`). Reword to "statements verified individually; not yet run top to bottom" until the Run All exists.
- **Lineage — `SemanticCatalog` survives** in notebook cell 0 (JSON line 17, presented as the current name), cell 23 and the test docstring line 2, while README:22 says `DataLineage`; the SDK CHANGELOG places the rename in **v4.1.1 (2026-08-31)**, marked breaking, not "by v4.2.1".
- **Lineage — README:81 "test_B0 prints the full candidate matrix on every run … you see it"**: B0 always passes, so with every documented invocation (`-v`, `-m existence`, `-rX`) the matrix is captured and never shown (executed on the real module); only `-s`/`-rP`/`-rA` show it, and the line-286 "NOW ACCEPTED" tripwire is unreachable by design. Also README:98 and docstring line 23: `-rX` prints nothing in the current all-xfail state; `-rx` shows the BLOCKED reasons. Document `-rP -k B0` or surface the state via `record_property`/`warnings.warn`.
- **Lineage — README:44 "If you probe and see 404, you are on the wrong generation"** is contradicted by your own #107 account: a wrong or unauthorised `AIDP_DATALAKE` yields the same `NotAuthorizedOrNotFound` 404 on the correct host (the module's own comment at 39–41 says so; A2 fires it on `datalake.{region}`). Executed with a mocked 404: A1/A3/A7 print "route missing", A2 passes, and the reader is sent to the generation section though host and version are hard-coded. Retitle to "Reading a 404", say what it means on each host, and point at A0 as the disambiguator.
- **Lineage — investigation voice remains** in shipped output: the BLOCKED xfail reason ends "Open question for Oracle." (printed by `-rx`/`-ra`), B0's docstring "the evidence to hand Oracle", the module docstring's "in this tenancy … our own tables", README:54–55 "so nobody re-derives the wrong conclusion". No data-engineering README on main addresses Oracle as an external party. Neutral rewording suggested per line.
- **Lineage — the negative finding now leads the README undated.** #107's TAKEAWAYS carried the observation date (2026-08-16); the #107 review asked to pin it to a date/SDK version; #116 promotes the gap to the opening callout (README:10–14) with only the SDK version (line 218) and tells every reader categorically "you will not get a lineage graph out of it" on the strength of one DataLake. Add "Observed 2026-08-16 on one DataLake in us-ashburn-1, SDK v4.2.1" to the callout, cell 0 and the BLOCKED constant, and soften 13–14 to "on that tenancy no graph came back; Part B reports XPASS if yours differs". The README will otherwise rot silently the day population lands.
- **Lineage — harness fragility carried from #107**: exact Preview-API message pins (A4/A5 fail on a message that merely echoes the value — executed 6 of 12 red), A7's dependency on the legacy gateway serving `/catalogs`, 60 s timeout with no retry (note `HTTPAdapter(max_retries=…)` does not retry POST read-timeouts), the OCID candidate printed in B0's `-s` output, `pytestmark = []`, `AIDP_DATALAKE` missing from the docstring's Run block. Capture the unresolvable sentinel's message once in a fixture and compare against it; give A7 its own marker.
- **Process — the body describes deltas versus #107, not main**: "Four files change, two are added, one is removed" (main-relative: 3 modified, 7 added, 0 removed), "29 cells to 26" (main has 25), "overwriteSchema removed" (main never had it), "`lake.scd` … three cells failed" and "%pip removed" (both #107-only; main targets `hive.default.dim_customer`). The same framing is baked into commit messages `f00ac82` ("bronze and silver writes failed on a type mismatch" — silver failed on the extra column, as your own cell 6 says) and `7d35533` ("Drops TAKEAWAYS.md" in an add-only commit). Reword if you rebase to strip the metadata; a squash-merge with a corrected message also cures it. Two real fixes vs main are undocumented: the `float()` wraps and `processed_timestamp` population.
- **Process — issues and #107.** #112 is open with no maintainer reply and your own text says "do not merge before this is settled" — that is the current merge gate for the lineage commit; "Relates to #113" should be "Fixes #113" so it closes on merge; nothing tracks the generator-vs-DDL defect in the six other medallion demos your note defers — please open one issue for it. CONTRIBUTING's directory list is not updated for `lineage/` (precedent: it has been stale since March, so note-level).

### Notes

- Cell 6: "Two distinct things can go wrong" introduces three bullets. `primary_service_type` picks an arbitrary winner on ties (`ROW_NUMBER … ORDER BY COUNT(*) DESC`, no tie-breaker; ≈13% of subscribers at 10k) — pre-existing; add `, service_type`. DECIMAL overflow is a silent NULL under the default ANSI setting and nothing checks the gold tables after the write; the widths are sized for 10,000 subscribers with ≈7.6× headroom — one sentence in cell 6, or a NULL assert after cell 18.
- SCD: same-day changes yield zero-length intervals and `current_date()` is evaluated in two statements (midnight straddle → end ≠ start); the step-2 tracked-column terms are never decisive after step 1 (fine, say why they are kept); the pair is two Delta commits, re-running recovers; source deletions are ignored (`WHEN NOT MATCHED BY SOURCE` parses on Delta 3.2 if wanted). Eight of the 12 SCD cells lost their nbformat 4.5 `id` (main had 10/10); the clone notebook has none of 17 — `nbformat.normalize` fixes both.
- Clone: cell 12 says "filtered with the DataFrame API" but applies no filter; the plain `DESCRIBE` subquery is an analysis error for a bare name (the qualified forms are parse errors); the CTAS has no `USING delta` while cells 2 and 14 do (on stock defaults it would be a Hive text table); the VACUUM warning omits the documented repair — and a bare re-`CREATE OR REPLACE … SHALLOW CLONE` over a non-empty clone throws `DELTA_UNSUPPORTED_NON_EMPTY_CLONE`, so say "`DELETE FROM` the clone, then re-clone"; cell 16 leaves the session on a dropped schema; re-running cell 6 alone prints 4 against "expect 3".
- Lineage: `(literal)` in cell 14 conflates constants with unresolved intermediate attributes (rename to "(no source columns resolved)"); the `hasattr()` guard in `_output_exprs` is always true on py4j objects (dispatch on `node` instead); `pytest.ini` is ignored when pytest runs from the repo root (13 unknown-mark warnings) — a 3-line `conftest.py` is not; the unset-variable stop is a pytest collection error, exit 2 — keep it (a session fixture + `pytest.fail` would be absorbed by the xfail marker, which is exactly the trap README:108 warns about), just say so in README:86/103; "Clean run: 19/19" and "13 passed / 4 xfailed" are evidenced only by #107's now-cleared outputs — date them.

### On the split question

Your #107 question ("split into four PRs, or fix in place?") went unanswered; Lucas's #107 review recommended four PRs. Fix-in-place with four single-concern commits works for me as a reviewer, with one caveat: the lineage commit is gated on #112, so if that stalls, peel `f00ac82` + `54b85d1` into a fixes-only PR rather than hold two verified bug fixes hostage. That is my view, not a maintainer decision. Either way, please close #107 as superseded.

**Bottom line:** the medallion and SCD repairs are done and verified; the lineage sample needs the anchor decoupling (major 2), the location redaction (major 3) and the scope wording (major 4), plus the naming/voice/date minors; the clone sample needs the LIKE/USE pattern replaced or an AIDP Run All pasted (major 1), real asserts or softer wording, and the location dropped from cell 5; strip the stale cell metadata and fix the body/commit framing. Happy to re-verify. 🤖 Deep review by Claude (Fable)

navid-heydari:
## All four majors fixed in `7451336` — the anchor coupling is reproduced and the fix verified in both states

@ahmedawan-oracle — all four majors are addressed in one commit on top of `f2786bc`. I verified each finding against the files before changing anything rather than taking them as given; all four hold. Details per item, including the one place I reached a different conclusion about the underlying cause than the suggested fix implied.

Scope note: this commit is the **four majors only**. The minors and notes are not in it — see the bottom for how I plan to handle them.

### 1. Clone sample — `USE` + `CREATE TABLE … LIKE`

Taken as correct and fixed the way you suggested. The `USE` is gone, every statement carries a three-part name via an `NS = f"{CATALOG}.{DB}"` prefix, and the `LIKE` example is replaced by the CTAS with an explicit `USING delta`. I did not paste an AIDP `DESCRIBE DETAIL` — I still have no workspace to Run All in, and your `SessionCatalog` / `setCurrentNamespace` reasoning means the pattern is wrong on upstream Spark regardless of what AIDP happens to do with it. A sample that only works on one catalog implementation is not worth keeping when the qualified form works everywhere.

`README.md:32` no longer teaches `USE` as the workaround; it now states why the pattern is avoided, with the `CreateTableLikeCommand` / v1 `SessionCatalog` reason and the silent-bind-to-`default` failure mode. Cell 8's sentence is rewritten to match. Verified mechanically that no bare table reference and no `USE` survive anywhere in the notebook.

### 2. Lineage harness — Part A asserting 400 on Part B's anchor

Confirmed, and I reproduced it before fixing it. I built an in-process mock of the data-plane implementing only the behaviours the module and README document — missing vs. invalid `anchorNode`, enum validation, the bogus-action 404, the legacy-host 404 — with a switch for whether the graph is populated.

| | today (unpopulated) | populated |
|---|---|---|
| `f2786bc` | 13 passed / 4 xfailed | **7 failed / 6 passed / 4 xpassed** |
| `7451336` | 13 passed / 4 xfailed | **13 passed / 4 xpassed** |

The populated-state failures on the old code are exactly the set you named: `A1`, `A3`, and all five `A5` parametrisations. That the numbers match your independent run is decent evidence both mocks are faithful.

Fixed as suggested: `fetch_lineage()` now defaults to `UNRESOLVABLE_ANCHOR`, `A3`'s `exportLineage` body uses the sentinel, and `B1`–`B3` pass `anchorNode=ANCHOR_TABLE` explicitly. One deviation — `A4` was hand-rolling its own `"definitely-not-a-node"` string; I pointed it at the shared sentinel so there is a single definition of "anchor that must never resolve" rather than two that could drift. The docstrings for Part A, `A1` and `A5` now say why the sentinel is used, so the next person does not helpfully "fix" it back to `ANCHOR_TABLE`.

`AIDP_ANCHOR_TABLE` is now safe to set: it only feeds Part B.

### 3. The location print

Confirmed — cell 18 was byte-identical to #107 and was indeed the source of the `oci://<bucket>@<namespace>` lines. Fixed with your suggested form: the full location still feeds the `startswith` check, and only the path is printed.

```python
print("     loc: .../%s" % loc.split("/", 3)[-1])
```

Checked against `oci://`, a short `oci://`, and an `s3a://` location — the authority is dropped in all three and the check is unaffected. The clone notebook no longer projects `location` from `DESCRIBE DETAIL`, with a comment saying why so it does not get added back. Both READMEs now carry a "clear outputs before committing" paragraph.

### 4. The narrowed scope claim

Confirmed, and your restatement is the right one. I did not need to run Spark to establish it — it falls out of cell 4 by inspection: `colmap` is populated exclusively from `collectLeaves()`, so any attribute minted by an intermediate `Project` has an `exprId` that is not a key in that map, and `ups` comes back empty. The top node's type is simply not the governing condition.

README and cell 23 now state the real condition — *every attribute the top node references is emitted directly by a leaf relation* — with `SELECT * FROM (SELECT cust_id, amount * 2 AS amt FROM raw_orders) s` as the worked counterexample, and a note that `.withColumn().withColumn()` fails the same way. New Scope row:

> | Alias computed in a `FROM`-subquery, or by an earlier `.withColumn()` / `.select()` — i.e. `Project` over `Project`, `Aggregate` over `Project` | that column's lineage is empty, printed as `(literal)` |

I went with rewording rather than the ~15-line transitive resolver. Adding it would widen the sample's claim right after narrowing it, and it would need its own negative controls to be worth trusting. Happy to do it as a follow-up if you would rather the extractor actually handled the shape.

### Verification

- Both notebooks parse as nbformat JSON, every code cell byte-compiles, outputs still cleared.
- No `oci://…@…` and no OCID in any file touched by this PR.
- Harness: 13 passed / 4 xfailed unpopulated, 13 passed / 4 xpassed populated; unset `AIDP_DATALAKE` still stops the suite with the named-variable message.
- Still no Run All on either notebook, for the same reason as before. Majors 1 and 3 reduce what a future Run All can leak, but they do not substitute for one.

### On the minors, the notes, and the split

Not in this commit, deliberately — I wanted the four blocking items reviewable on their own rather than buried in a diff that also strips cell metadata across 19 cells.

On your split recommendation: agreed, and I would like to take it now. #112 has no maintainer answer and it gates the lineage commit, so holding `f00ac82` + `54b85d1` behind it is the wrong trade. Unless you object I will peel the two verified notebook fixes into a fixes-only PR and leave this one for the two new samples, then work the minors here while #112 settles.

Three of your items need a decision rather than a patch, so flagging them rather than guessing:

- **Stale cell metadata** (115 of 298 added lines) — I will strip `command_metadata` / `execution` / `result_type` and keep `kernelspec` / `language_info`, matching `main`. Worth a repo-level `nbstripout` guard in CONTRIBUTING so this stops recurring; say the word and I will open an issue.
- **Body and commit-message framing describing deltas vs. #107 rather than `main`** — cleanest cure is a squash-merge with a corrected message. If you would rather I rebase and reword the four commits, I will.
- **The generator-vs-DDL defect in the other six medallion demos** — nothing tracks it. I will open that issue.

I will also close #107 as superseded.

navid-heydari:
1. Majors were addressed. 
2. Working on remaining Minors item captured by Ahmed.

navid-heydari:
## Minors and notes done in `6e3828c`; PR body rewritten to be `main`-relative

@ahmedawan-oracle — follow-up to `7451336`. Everything from the Minors and Notes sections is in `6e3828c`, and the body framing item is fixed by rewriting the description rather than rebasing. Two places I concluded something different from the review are called out below.

### Medallion

- **Stale run record stripped.** Confirmed exactly as described: `command_metadata`/`execution`/`result_type` on all 19 code cells, `result_type: "error"` on 5, 11 and 24, and `Last_Active_Cell_Index: 27` on a 26-cell notebook. `main` has zero cell metadata here. The diff for this file is now **+44/−201** instead of the previous +298.
- **One correction.** I initially stripped cell metadata wholesale and that was wrong: `type` is a workbench convention, not run noise — `main`'s own SCD notebook carries it, and 30 of 155 notebooks on `main` use it. Only `execution`/`command_metadata`/`result_type` are stripped now; `type` is preserved, and restored on the SCD notebook where I had over-stripped it.
- **`DROP` added before bronze and silver.** Your reasoning applies to them and I had only applied it to gold.
- **The three gold creates are now plain `CREATE TABLE`.** That was the cleaner half of the contradiction you found — the comment says "Dropped first, not `CREATE TABLE IF NOT EXISTS`" and was followed by exactly that statement, so the statement changed rather than the comment.
- **Wildcard import replaced** with explicit names plus `round as spark_round`, so pyspark's `round`/`max`/`min` no longer shadow the builtins the generator cell calls on plain numbers.
- Tie-breaker `, service_type` added; the silent-NULL-on-overflow behaviour and the 10,000-subscriber sizing assumption are now stated in the RCA cell; "two distinct things" over three bullets fixed.

### SCD

- **The duplicate-key precondition is now stated both ways and enforced.** Your distinction is the important part — the new-key case is not an error, it is silent corruption — so `run_scd2_merge` raises before executing either statement rather than only documenting the requirement. Verified: the guard fires with 0 statements executed and names the offending keys, and the expired `[1, 4]` / inserted `[1, 3, 4]` behaviour is unchanged by it.
- Same-day zero-length intervals, the two separate `current_date()` evaluations, the two-commit non-atomicity, and ignored source deletions (with the `WHEN NOT MATCHED BY SOURCE` pointer) are all documented. The step-2 tracked-column terms now say why they are kept.
- The 8 missing nbformat 4.5 `id`s are restored and `Last_Active_Cell_Index` is gone.

### Clone

- **Real `assert`s per step.** The body's claim was false and is now true rather than softened.
- **You were right that `DESCRIBE DETAIL` never demonstrated zero-copy.** The notebook now asserts `operationMetrics.numCopiedFiles == 0` on the `CLONE` commit *and* that the clone's `inputFiles()` are exactly the source's, then shows them diverging after the insert. That is the actual proof; the previous cell would have looked identical for a deep copy.
- Divergence cell made idempotent (it printed 4 against "expect 3" when re-run alone), VACUUM repair path documented including `DELTA_UNSUPPORTED_NON_EMPTY_CLONE`, "Environment as tested" reworded to "verified individually; not yet run top to bottom", 17 missing cell `id`s added.

### Lineage

- **`pytest.ini` → `conftest.py`**, and I verified the problem before changing it: from the repository root the ini file yields unknown-mark warnings and the `conftest.py` yields none. It also registers the new `legacy` marker.
- **Message pins removed.** `A1`/`A4`/`A5`/`A6` no longer pin the literal `"Invalid anchorNode"`; a session fixture captures the server's own wording from one probe and the rest compare against that. The discrimination those tests are actually about — reached anchor resolution vs. rejected earlier — is preserved without pinning Preview-API wording.
- **`A7` marked `legacy`** so `-m "existence and not legacy"` keeps the other 11 Part A tests when the old gateway goes away. Verified: 11 passed, 6 deselected.
- **OCID redacted from `B0`'s output.** Verified the OCID value appears nowhere in full `-rA` output.
- Dead `pytestmark = []` removed; the no-retry decision is recorded with the reason (`HTTPAdapter(max_retries=...)` does not retry a POST that times out reading the response).
- **`_output_exprs` now dispatches on the node type.** Worth being precise about this one: the `hasattr()` guard was dead, as you said, but the function still produced correct results — a `Project` accessor call on an `Aggregate` raises through py4j and the `except Exception: pass` fell through to the right branch. So this is a clarity and fragility fix, not a behaviour fix. Verified against a stub whose `__getattr__` synthesises every name: `Aggregate` → `aggregateExpressions`, `Project` → `projectList`, and `Sort`/`WithCTE`/`Filter` → `None`.
- `(literal)` renamed to `(no source columns resolved)`; service name corrected to `DataLineage` in the notebook; rename dated to **v4.1.1 (2026-08-31)** where it said "by v4.2.1".
- **README:** the negative finding is dated (2026-08-16, one DataLake in us-ashburn-1, SDK v4.2.1) in the opening callout and softened to what one observation supports; the 404 section is retitled "Reading a 404" with a per-host table and points at `A0` as the disambiguator; `B0`'s matrix now documents `-rP -k B0` and `-rx` replaces `-rX`; the collection-error-by-design decision is stated; the remaining investigation voice is gone.
- `data-engineering/lineage/` added to CONTRIBUTING's directory list.

Regression check after all of the above: **13 passed / 4 xfailed** unpopulated, **13 passed / 4 xpassed** populated, zero unknown-mark warnings, all four notebooks nbformat-valid with every code cell compiling and outputs cleared.

### Body framing

Rewritten rather than rebased, so the description now describes `main`: **4 files modified, 7 added, 0 removed**; medallion 25 → 26 cells; SCD 10 → 12. The claims that were `#107`-relative are gone — `main` never had `overwriteSchema`, never had `%pip install jinja2`, and targets `hive.default.dim_customer` rather than `lake.scd`. Both undocumented fixes are now in the body: `processed_timestamp` was declared in `main`'s silver DDL and never populated, and the `float()` wraps are a consequence of gold honouring its `DECIMAL` DDL (`Decimal * float` raises). `Relates to #113` is now `Fixes #113`.
### Closed
- #107 was closed as reported in morning.
### Still open

- **#112** — unanswered, and it gates the lineage commit only. My preference is on the record in the body: if it stalls, peel `f00ac82` + `54b85d1` into a fixes-only PR.
- **The other medallion demos.** I confirmed your count independently: **6 of the other 11** declare narrow bronze types and write an uncast `createDataFrame` frame — `energy`, `financial_services`, `manufacturing`, `real_estate`, `retail`, `transportation`. `insurance` already casts; `education`, `healthcare`, `hospitality`, `media` declare no narrow bronze types. I have the issue text drafted and will file it.

PR #117, opened 2026-09-21
Update Ask-AIDP to 0.10.0 with September CLI and REST coverage

vishaldhiman22:
## Summary
Update the Codex Ask-AIDP plugin at ai/codex-plugins/plugins/ask-aidp and its parent marketplace README, based on the latest upstream main.

- Refresh the Oracle CLI reference to 256 commands across 18 groups and the REST reference to 271 operations across 19 categories.
- Add reference coverage for 14 new CLI commands: data lineage export/fetch, bundle publishing/status, compute cloning and configuration import/export/get, Maven search, workspace and volume ZIP upload/download, and task-run retries.
- Update skill guidance, examples, version metadata, and changelog while preserving the upload, CSV ingestion, installation, and packaging fixes from 0.9.1.
- Test lookups and REST request planning for all 14 additions, including the documented POST method for bundle publish status.
- Preserve relative dependency symlinks in packaged archives so they do not point back to the build machine.

## Validation
- Source QA passed with Oracle aidp-cli 4.2.1.
- All 256 catalog commands matched the installed CLI operation manifest.
- Plugin manifest and skill validation passed; JavaScript syntax and git diff checks passed.
- Built tar.gz and ZIP archives with aidp-cli and aidp-typescript-client 4.2.1.
- Extracted tar.gz QA passed using its bundled CLI, including portability checks.
- Existing 43 MCP tools retained.

Live Oracle AI Data Platform operations were not run. REST dry runs validate request routing, not server-side payload acceptance. Generated archives and dependencies remain outside the source repository, following its existing distribution convention.


craxelfn:
The new `cluster export-compute-configuration` entry appears to expose a nested model as the top-level request schema.

In `assets/aidp-cli-command-reference.json:3389`, `bodyModel` is `ComputeConfigurationLibraryEntry`, which contains only `jar`, `path`, and `pip`. Calling `aidp_cli_reference` with default arguments reproduces this output. However, the entry’s full Markdown includes `ExportComputeConfigurationDetails`, with `clusterScopedLibraries`, `environmentVariables`, `destinationPath`, and `fileName`. An agent relying on the default lookup could therefore construct an incorrect export payload.

I attempted live validation on two AIDP instances. Export POSTs initially returned HTTP 406 because the generic REST tool defaults to `Accept: application/json`. Setting `Accept: application/x-yaml` resolved that rejection, but both the nested and catalog-shaped payloads then returned HTTP 403: Compute Configuration is not enabled for this AIDP. This confirms the response-header issue, while the runtime impact of the schema discrepancy remains unverified.

Could we verify the intended root request model, update the generator/catalog accordingly, and ensure the export flow supplies the appropriate response header? The new QA checks validate command lookup and routing, but should also assert the returned body model, root fields, and effective headers. Once an enabled instance is available, a successful export and YAML readback would complete the validation.

vishaldhiman22:
Addressed in commit 6349aa2.

Confirmed the request contract against the [Oracle export operation reference](https://docs.oracle.com/en/cloud/paas/ai-data-platform/aiwap/op-aidataplatforms-aidataplatformid-workspaces-workspacekey-clusters-clusterkey-actions-exportcomputeconfiguration-post.html), `aidp-cli` 4.2.1's operation manifest, and the TypeScript SDK request serializer: the root is `ExportComputeConfigurationDetails`, with `clusterScopedLibraries`, `environmentVariables`, `destinationPath`, and `fileName`. `ComputeConfigurationLibraryEntry` is a nested library item.

Changes:
- The generator selects the root model from the CLI manifest rather than the first Markdown model. It rejects ambiguous/mismatched inputs and preserves documented JSON field spelling. Regeneration corrected or filled in 44 model entries; all 256 commands' root models and field names were checked against the manifest.
- Compute export REST POSTs default to `Accept: application/x-yaml`; JSON bodies get `Content-Type: application/json`. Headers are merged once because OCI's composer appends default and caller headers rather than overriding them. Dry-run output shows the effective headers; case-insensitive explicit overrides remain supported.
- Regression tests assert the default MCP lookup's root model and four fields, parser ordering and failure cases, composed headers, override behavior, and mocked YAML response preservation. Documentation includes the remaining live acceptance procedure.

Validation passed: full offline `scripts/qa.mjs`, plugin/skill validators, MCP self-test, catalog generation, and a separate transport check using the real `oci-common` 2.126.3 composer with mocked HTTP. No service call was made for these tests.

Live validation remains pending: the reported feature-disabled 403 responses do not establish payload correctness. We still need an enabled instance for a successful export followed by retrieval and comparison of the created workspace YAML. This update does not claim that acceptance test has passed.

